PR #1, opened 2018-11-08
Fix `doHaddockHydra` for doc output changes

jbgi:
btw, could this be a bit simpler with `overrideCabal`?


michaelpj:
Note that this is only compatible with 18.09, so if someone tries to use this with an earlier version of nixpkgs it won't work. The directory structure of the outputs changed between the two. I don't know what if anything we should do about that (maybe we're okay only supporting 18.09?).

disassembler:
nice catch! I'm fine with just supporting 18.09. I don't see a reason to support older nix versions. And since we pin our nix versions instead of use the host NIX_PATH we don't have to worry about breakage.

PR #2, opened 2018-11-09
Add pinned nixpkgs to common nix repo

rvl:
If there are multiple cardano repos it would make sense to have `iohk-nix` provide the pinned nixpkgs set -- as well as library functions depending on nixpkgs.

### DIFF
--- a/cardano-sl/nixpkgs-src.json
+++ b/cardano-sl/nixpkgs-src.json
@@ -1,0 +1,5 @@
+{
+  "url": "https://github.com/input-output-hk/nixpkgs",
+  "rev": "d4cd290a9a1b2ee7ac67ca87d5ffc2f3941157c0",
+  "sha256": "1kxkhccvwc3m9wdk9p57bafzgplk68j5l3rh7bkg6hdrzslyw4mh"
+}
--- a/clean-source-haskell.nix
+++ b/clean-source-haskell.nix
@@ -1,9 +1,8 @@
-pkgs: src:
-  let lib = pkgs.lib;
-  in if (builtins.typeOf src) == "path"
+{ lib }: src:
+  if (builtins.typeOf src) == "path"
     then lib.cleanSourceWith {
-      filter = with pkgs.stdenv;
-        name: type: let baseName = baseNameOf (toString name); in ! (
+      filter = name: type:
+        let baseName = baseNameOf (toString name); in ! (
           # Filter out cabal build products.
           baseName == "dist" || baseName == "dist-newstyle" ||
           baseName == "cabal.project.local" ||
--- a/daedalus/nixpkgs-src.json
+++ b/daedalus/nixpkgs-src.json
@@ -1,0 +1,5 @@
+{
+  "url": "https://github.com/input-output-hk/nixpkgs",
+  "rev": "fda4b93cd4fd3775408117c380aab0f33737d30f",
+  "sha256": "0f3nnsqgr1qpy9h65ssfmwj6c5nxadk3p043qca4glzw2k5pyrpw"
+}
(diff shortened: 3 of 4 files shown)


jbgi:
I feel it's a bit weird to pin a specific nixpkgs revision for every project we have in this repo.
Don't we rather stay completely project agnostic in this repo?
I think having one default pinned nixpkgs repo (maybe one for each nixpkgs channel we support) would be enough... what do you think?

michaelpj:
This project is implicitly dependent on a version of nixpkgs since it e.g. makes assumptions about the way the haskell packages are structured, their attributes, outputs, etc. So I don't think it would be crazy to put the pin here.

I would also appreciate being sure that I was going to naturally stay (roughly) in sync with the other IOHK repos in terms of our nixpkgs dependencies.

The danger would be situations where you want to update to a newer version of `iohk-nix` but that forces an undesirable nixpkgs bump. Not sure how big an issue that is.

jbgi:
> This project is implicitly dependent on a version of nixpkgs since it e.g. makes assumptions about the way the haskell packages are structured, their attributes, outputs, etc. So I don't think it would be crazy to put the pin here.

Agreed: it's reasonnable that this repo pin a default, compatible, revision of nixpkgs (overridable of course). What I would rather avoid is pinning a specific (potentially different) revision for each project (`plutus`, `cardano-sl`, `daedalus`...) in this repo.

michaelpj:
How about this: this repo has a "preferred" nixpkgs pin which is defined here, but can be overridden by other repositories that use it.

That way the "default" case is everyone on the same pin, but you can deviate if you need to.

rvl:
Yes it would be nice to only have one nixpkgs version that everything shares but if we do this we are making things unnecessarily difficult for ourselves.

For example, if you needed to upgrade nixpkgs for cardano-sl from 18.03→18.09 to fix a bug. If the package set is shared then we have introduced a dependency on making daedalus also work with 18.09 before we can fix the bug in cardano-sl.

So "what Michael said" -- these common expression can provide the preferred package set, but there is also a function for customizing it.

The goal of this is to help make sure `cardano-*` repos are using a common nixpkgs but letting them deviate if necessary.

michaelpj:
I'm really not a fan of having things referencing the downstream repos in here.

I think the difference with what I suggested is that I was expecting that overriding the common nixpkgs would happen *in the repo that needs it*. i.e. when you import `<iohk_nix>`. As it is, this scales pretty poorly as we add more projects to this, whereas putting it in the client repository seems easy enough (similar to what we're doing now). In the happy case where you don't actually care that much (like plutus) we can just not have a pin in our repository and use the blessed `iohk_nix` one.

rvl:
I don't envisage having a separate attribute with pinned nixpkgs for every single downstream repo, because that _would_ scale poorly.

Rather the `cardano-sl.pkgs` version of nixpkgs would be shared by all `cardano-*` repos. I want to avoid having to keep both `nixpkgs-src.json` (plus overlays) and `iohk-nix-src.json` in sync in every cardano repo.

Daedalus being a nodejs project has its own nixpkgs set. That could be kept in the client repository. But if we want to share common nix code which depends on `pkgs` then I think it's easier to keep the pinned nixpkgs for Daedalus inside iohk-nix.

Basically I think we could have about 2 or 3 "blessed" nixpkgs versions (less would be better though). Perhaps we could rename the attributes from cardano-sl/daedalus/plutus to something less specific?

michaelpj:
Ah I see - you have a *group* of repositories who you want to have a shared pin, but where you don't think you can count on always being able to use the `iohk_nix` pin.

Sounds like you need `cardano_nix` :) I can't think of any logical way of naming them if we keep them in here, since it sounds like this is inherently about managing the integration of a set of downstream repos.

PR #9, opened 2018-11-14
Add pinned nixpkgs to common nix repo

disassembler:
This needs more discussion. It completely break plutus currently which is the only package I think currently using iohk-nix.

### DIFF
--- a/cardano-sl/nixpkgs-src.json
+++ b/cardano-sl/nixpkgs-src.json
@@ -1,0 +1,5 @@
+{
+  "url": "https://github.com/input-output-hk/nixpkgs",
+  "rev": "d4cd290a9a1b2ee7ac67ca87d5ffc2f3941157c0",
+  "sha256": "1kxkhccvwc3m9wdk9p57bafzgplk68j5l3rh7bkg6hdrzslyw4mh"
+}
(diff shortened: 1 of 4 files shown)


michaelpj:
Here's a straw proposal:

# Version 1

`iohk_nix` takes a `pkgs` parameter, and if it is not set it evaluates a pinned nixpkgs in the `iohk-nix` repository and uses that. This argument is passed through to everything else and made available in the `iohk_nix` attribute set.

If a project wants to override that, e.g. because they need to upgrade nixpkgs faster than `iohk_nix` or they need to upgrade `iohk_nix` and not upgrade their pkgs, then they can pass in a pinned nixpkgs when they import it.

But what about `fetchNixpkgs`? Didn't we want to share that too?

# Version 2

As version 1, except `iohk_nix` is an attribute set with two attributes, one of which is `fetchNixpkgs` and one of which is the attribute set above (which expects to be passed a package set). So you can import `iohk_nix`, use `fetchNixpkgs` to get a pin, and then pass it to the rest.

Maybe you could do this more simply with some clever use of overrides, I don't know.

rvl:
This is how I am using iohk-nix in cardano-shell -- see `default.nix` in input-output-hk/cardano-shell#11

PR #10, opened 2018-11-15
common-nixpkgs improvements

disassembler:


disassembler:
I'm still not a fan of application specific stuff in iohk-nix, but I think this might be a compromise we can work with @rvl @michaelpj.

PR #13, opened 2018-11-20
Profiling: fix disabling profiling

### LINKED ISSUE #12: Disabling profiling isn't working
Not sure why not, possibly my fault. @disassembler

michaelpj:
The nixpkgs default is to enable library profiling, so turning off the
overlay won't disable profiling, we need to explicitly do it.

Also restructured the overlay a bit to be very careful about avoiding
rebuilds.

This fails for me with a linker error when trying to build the test
suite for `language-plutus-core` with profiling enabled, but that was
already true when profiling was enabled, so that's a separate issue.

Fixes #12.

### DIFF
--- a/haskell-packages.nix
+++ b/haskell-packages.nix
@@ -36,7 +36,7 @@ let
       then filterOverrides.benchmark else filter;
   };
   profilingOverlay    = import ./overlays/profile.nix       {
-    inherit pkgs;
+    inherit pkgs enableProfiling;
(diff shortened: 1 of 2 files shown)


michaelpj:
For clarity: I tested this both with and without profiling enabled, and observed:
- With profiling, my libraries and executables were being built with profiling.
- Without profiling they were not.
- There were no rebuilds of dependencies in either case, which is correct as they are built with profiling by default, which makes them suitable in either case.

PR #17, opened 2018-12-13
Add stack2nix regeneration script

michaelpj:
Also tweaked to be very similar to the nix-tools version. One difference
(which might be an improvement for the nix-tools script too) is taking
the output location as an argument, since we probably don't want to be
too prescriptive about that.

The hackage snapshot remains as an argument to the nix derivation so that we can
make a pinned regeneration script in a project's `default.nix`, since I would argue
that this is a slightly more important piece of data and should appear
with the rest of the nix configuration rather than be hidden in a script
argument.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -44,6 +44,7 @@ let
     haskellPackages = import ./haskell-packages.nix;
     commitIdFromGitRepo = pkgs.callPackage ./commit-id.nix {};
   };
+
   nix-tools = rec {
     # Programs for generating nix haskell package sets from cabal and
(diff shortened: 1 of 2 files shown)


angerman:
I'm afraid this will become obsolete with the nix-toolification that's currently happening. Thoughts?

michaelpj:
Well, it's functional. If nobody is using it after a while we can delete it, but at the moment it's a technique we use that isn't Plutus-specific, so it feels like it belongs here.

angerman:
Aye.

PR #20, opened 2019-01-18
Thread crossSystem through

angerman:
This is essential to have any hope of getting cross compilation to
work with iohk-nix.


jbgi:
@angerman I see now that there was a similar change in https://github.com/input-output-hk/iohk-nix/pull/16

PR #21, opened 2019-01-18
Bump nixpkgs to include ghc-8.6.3

angerman:
This probably conflicts with #18. As we really only care for the compilers. I wonder if it would make our life easier if we had the compilers in our own repository. They could happily live next to haskell.nix.

### DIFF
--- a/nixpkgs-pins/default-nixpkgs-src.json
+++ b/nixpkgs-pins/default-nixpkgs-src.json
@@ -1,6 +1,6 @@
 {
   "url": "https://github.com/input-output-hk/nixpkgs",
-  "rev": "ce68b263fe25a05a1111acac8ccbe538d139d2dc",
-  "sha256": "0b0r5zz4qfbr8x5gi60lwdxpnr670mk49idldh6sh66z315hmb0p",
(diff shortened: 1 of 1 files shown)


angerman:
Also we could get rid of that awful nix copy pasta that the ghc derivations are. There are sparate expressions for each ghc version. Not only does this mean you are limited to the ghc versions nixpkgs maintainers thing are useful, but it also means that there can be (and there were!) subtle differences between the expressions.

Cf. https://github.com/NixOS/nixpkgs/tree/bb1cb2a332d75c4065d90583270a8ef0172762e9/pkgs/development/compilers/ghc

angerman:
@michaelpj what's your take on this? (Hey, let me rope you more into nix :p)

michaelpj:
Hm, I thought we were sticking to the stable branch? Can't we just update the `iohk-18.09` branch? Or has 8.6.3 not been backported to 18.09?

michaelpj:
(I certainly approve of a bump. Also, since it's unclear it would also be good to write down some policy on updating the pin, maybe in the README)

angerman:
@michaelpj yea, I've come to the realization that I'd rather back port ghc-8.6 to our nixpkgs fork from the stable branch instead of going into unchartered unstable waters.

PR #16, opened 2018-12-13
[DEVOPS-1179] Updates to common nix-tools for cardano-wallet

rvl:
More changes for cardano-wallet nix-tools build - input-output-hk/cardano-wallet#153.

PR #22, opened 2019-01-18
[DEVOPS-1129] nix-tools related fixes

jbgi:
super-seed #16.


angerman:
Just for reference: the current cardano-sl develop has the hash that we update for cardano-sl:

https://github.com/input-output-hk/cardano-sl/blob/e8595ce910677c74aba6aeb1bd2f0077a186b7ee/nixpkgs-src.json#L2-L5

Thus with respect to cardano-sl nothing changes.

PR #26, opened 2019-01-22
Fix version typo

michaelpj:
Seems the wrong way around.


michaelpj:
@angerman

michaelpj:
I'm only nagging because I want to experiment with 8.6.3 and not accidentally 8.6.2 ;)

PR #30, opened 2019-01-23
Moves nix-tools standard templates into iohk-nix

angerman:
As such we can now define:

```nix
let
  localLib = import ./lib.nix;
in
{ pkgs, ... }@args
localLib.nix-tools.default-nix ./nix/pkgs.nix args // {
    # additional non-haskell packages we want to build on CI.
    byronLedgerSpec = import ./specs/ledger/latex {};
    byronChainSpec = import ./specs/chain/latex {};
    semanticsSpec = import ./specs/semantics/latex {};
}
```
as `default.nix`

and

```nix
let
  localLib = import ./lib.nix;
in
localLib.nix-tools.release-nix {
  package-set-path = ./.;

  # packages from our stack.yaml or plan file (via nix/pkgs.nix) we
  # are intereted in building on CI via nix-tools.
  packages = [ "cardano-chain" "cs-ledger" "small-steps" "cs-blockchain" ];

  # The set of jobs we consider crutial for each CI run.
  # if a single one of these fails, the build will be marked
  # as failed.
  #
  # The names can be looked up on hydra when in doubt.
  #
  # custom jobs will follow their name as set forth in
  # other-packages.
  #
  # nix-tools packages withh be prefixed with nix-tools and
  # follow the following naming convention:
  #
  #   namespace                      optional cross compilation prefix                 build machine
  #   .-------.                              .-----------------.                .--------------------------.
  #   nix-tools.{libs,exes,tests,benchmarks}.{x86_64-pc-mingw-,}$pkg.$component.{x86_64-linux,x86_64-darwin}
  #             '--------------------------'                    '-------------'
  #                 component type                          cabal pkg and component*
  #
  # * note that for libs, $component is empty, as cabal only
  # provides a single library for packages right now.
  #
  # Example:
  #
  #   libs.cardano-chain.x86_64-darwin -- will build the cardano-chain library on and for macOS
  #   libs.cardano-chain.x86_64-linux -- will build the cardano-chain library on and for linux
  #   libs.x86_64-pc-mingw32-cardano-chain.x86_64-linux -- will build the cardano-chain library on linux for windows.
  #   tests.cs-ledger.ledger-delegation-test.x86_64-linux -- will build and run the ledger-delegation-test from the
  #                                                          cs-ledger package on linux.
  #
  required-name = "cardano-chain-required-checks";
  required-targets = jobs: [
    jobs.byronLedgerSpec
    jobs.byronChainSpec
    jobs.semanticsSpec

    jobs.nix-tools.libs.cs-blockchain.x86_64-darwin
    # ...

    # windows cross compilation targets
    jobs.nix-tools.libs.x86_64-pc-mingw32-cardano-chain.x86_64-linux
    # ...
  ];
}
```
as `release.nix`.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -60,6 +60,7 @@ let
     default-nix = import ./nix-tools-default.nix (commonLib // { inherit nix-tools; });
     release-nix = import ./nix-tools-release.nix (commonLib // { inherit nix-tools; });
     iohk-module = import ./nix-tools-iohk-module.nix commonLib;
+    iohk-overlay = import ./nix-tools-iohk-overlay.nix commonLib;
   };
 
   stack2nix = rec {
--- a/default.nix
+++ b/default.nix
@@ -55,6 +55,11 @@ let
     regeneratePackages = commonLib.pkgsDefault.callPackage ./nix-tools-regenerate.nix {
       nix-tools = package;
     };
+    # default and release templates that abstract
+    # over the details for CI.
+    default-nix = import ./nix-tools-default.nix (commonLib // { inherit nix-tools; });
+    release-nix = import ./nix-tools-release.nix (commonLib // { inherit nix-tools; });
+    iohk-module = import ./nix-tools-iohk-module.nix commonLib;
   };
 
   stack2nix = rec {
--- a/default.nix
+++ b/default.nix
@@ -59,7 +59,16 @@ let
     # over the details for CI.
     default-nix = import ./nix-tools-default.nix (commonLib // { inherit nix-tools; });
     release-nix = import ./nix-tools-release.nix (commonLib // { inherit nix-tools; });
+
+    # default iohk module and overlay to be used in the pkgs.nix file of the
+    # project.  The module will provide the necessary default overrides for
+    # packages (patches) to work properly in cross compiled settings.
     iohk-module = import ./nix-tools-iohk-module.nix commonLib;
+    # The overlay provides the necessary extra packages that might be missing
+    # from generated plans (mostly from stackage snapshot) as well as patches
+    # to align the packages downloaded from hackage with what GHC ships as those
+    # packages.  That a package of a given version on hackage is identical to
+    # the package that ghc ships with the same version is not a given!
     iohk-overlay = import ./nix-tools-iohk-overlay.nix commonLib;
   };
 
--- a/haskell-hackage-stackage.nix
+++ b/haskell-hackage-stackage.nix
@@ -1,0 +1,38 @@
+{ pkgs }:
+let
+  # overriding logic so we can pass -I to nix, and overide the
+  # relevant import.
+  overrideWith = override: default:
+   let
+     try = builtins.tryEval (builtins.findFile builtins.nixPath override);
+   in if try.success then
+     builtins.trace "using search host <${override}>" try.value
+   else
+     default;
+in rec {
+  # all packages from hackage as nix expressions
+  hackage = import (overrideWith "hackage"
+                    (pkgs.fetchFromGitHub { owner  = "angerman";
+                                            repo   = "hackage.nix";
+                                            rev    = "d4ef2536554d78c6adf368816a4476bc909fcd96";
+                                            sha256 = "15siby9iwgi8bhkj83q1djvg6b6n4gj08n7yy66ccjck7fw4hbr8";
+                                            name   = "hackage-exprs-source"; }))
+                   ;
+  # a different haskell infrastructure
+  haskell = import (overrideWith "haskell"
+                    (pkgs.fetchFromGitHub { owner  = "angerman";
+                                            repo   = "haskell.nix";
+                                            rev    = "ed456599a3a52592ffede765e9d12a953b77d606";
+                                            sha256 = "0w7dg50y30zm7k5ilw0mhkggzhmzpb7x8cl2b82iizvscx6xffzc";
+                                            name   = "haskell-lib-source"; }))
+                   hackage;
+
+  # the set of all stackage snapshots
+  stackage = import (overrideWith "stackage"
+                     (pkgs.fetchFromGitHub { owner  = "angerman";
+                                             repo   = "stackage.nix";
+                                             rev    = "f58d5b78e7a40260c6142c79e52c2bf3ae9876b9";
+                                             sha256 = "1nd9lfm016rlhw3133488f8v8x3lbxrld422gw8gcjhhfls3civn";
+                                             name   = "stackage-snapshot-source"; }))
+                   ;
+}
--- a/haskell-hackage-stackage.nix
+++ b/haskell-hackage-stackage.nix
@@ -12,7 +12,7 @@ let
 in rec {
   # all packages from hackage as nix expressions
   hackage = import (overrideWith "hackage"
-                    (pkgs.fetchFromGitHub { owner  = "angerman";
+                    (pkgs.fetchFromGitHub { owner  = "input-output-hk";
                                             repo   = "hackage.nix";
                                             rev    = "d4ef2536554d78c6adf368816a4476bc909fcd96";
                                             sha256 = "15siby9iwgi8bhkj83q1djvg6b6n4gj08n7yy66ccjck7fw4hbr8";
@@ -20,7 +20,7 @@ in rec {
                    ;
   # a different haskell infrastructure
   haskell = import (overrideWith "haskell"
-                    (pkgs.fetchFromGitHub { owner  = "angerman";
+                    (pkgs.fetchFromGitHub { owner  = "input-output-hk";
                                             repo   = "haskell.nix";
                                             rev    = "dae8025469d94e739aa52f23685bcd87840c72e3";
                                             sha256 = "0bj577aj2v7yl6d1l6rinp0bbn4k9m8k4jd46m4hspqaisfd58hg";
@@ -29,7 +29,7 @@ in rec {
 
   # the set of all stackage snapshots
   stackage = import (overrideWith "stackage"
-                     (pkgs.fetchFromGitHub { owner  = "angerman";
+                     (pkgs.fetchFromGitHub { owner  = "input-output-hk";
                                              repo   = "stackage.nix";
                                              rev    = "f58d5b78e7a40260c6142c79e52c2bf3ae9876b9";
                                              sha256 = "1nd9lfm016rlhw3133488f8v8x3lbxrld422gw8gcjhhfls3civn";
--- a/haskell-hackage-stackage.nix
+++ b/haskell-hackage-stackage.nix
@@ -22,8 +22,8 @@ in rec {
   haskell = import (overrideWith "haskell"
                     (pkgs.fetchFromGitHub { owner  = "angerman";
                                             repo   = "haskell.nix";
-                                            rev    = "ed456599a3a52592ffede765e9d12a953b77d606";
-                                            sha256 = "0w7dg50y30zm7k5ilw0mhkggzhmzpb7x8cl2b82iizvscx6xffzc";
+                                            rev    = "dae8025469d94e739aa52f23685bcd87840c72e3";
+                                            sha256 = "0bj577aj2v7yl6d1l6rinp0bbn4k9m8k4jd46m4hspqaisfd58hg";
                                             name   = "haskell-lib-source"; }))
                    hackage;
 
--- a/haskell-hackage-stackage.nix
+++ b/haskell-hackage-stackage.nix
@@ -1,7 +1,10 @@
 { pkgs }:
 let
   # overriding logic so we can pass -I to nix, and overide the
-  # relevant import.
+  # relevant import. That is if we want to override for example
+  # `hackage`, we can simply provide -I hackage=/path/to/hackage.nix
+  # and have nix use that instead of the one we reference here.
+  # Same for haskell and stackage.
   overrideWith = override: default:
    let
      try = builtins.tryEval (builtins.findFile builtins.nixPath override);
--- a/haskell-hackage-stackage.nix
+++ b/haskell-hackage-stackage.nix
@@ -14,8 +14,8 @@ in rec {
   hackage = import (overrideWith "hackage"
                     (pkgs.fetchFromGitHub { owner  = "input-output-hk";
                                             repo   = "hackage.nix";
-                                            rev    = "d4ef2536554d78c6adf368816a4476bc909fcd96";
-                                            sha256 = "15siby9iwgi8bhkj83q1djvg6b6n4gj08n7yy66ccjck7fw4hbr8";
+                                            rev    = "1bee48237cb87d9ad6432a9ddb7777f858674400";
+                                            sha256 = "0sllxmffrcvcq63wlvz64s0i0ls2249vwix645899k6xh159z5pj";
                                             name   = "hackage-exprs-source"; }))
                    ;
   # a different haskell infrastructure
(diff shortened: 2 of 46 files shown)


default.nix
```diff
@@ -55,6 +55,10 @@ let
     regeneratePackages = commonLib.pkgsDefault.callPackage ./nix-tools-regenerate.nix {
       nix-tools = package;
     };
+    # default and release templates that abstract
+    # over the details for CI.
+    default-nix = import ./nix-tools-default.nix commonLib;
+    release-nix = import ./nix-tools-release.nix commonLib;
```
michaelpj: Are these usable as part of a larger `default.nix` or `release.nix`? For example, in the Plutus repo we also have a bunch of latex that's buildable via our top-level `default.nix` and included in `release.nix`.

Also, writing a little about how to use these in a README or similar would be good.

nix-tools-default.nix
```diff
@@ -0,0 +1,37 @@
...
+let  nix-tools = import pkgs-path { nixpkgs = _: pkgs; };
+in {
+    _lib = commonLib;
+
+    # This will allow us to build
+    # nix-tools.libs.cardano-chain to obtain all libs in a single derivation
+    # nix-tools.exes.cardano-chain for all executables, same for tests and benchmarks.
+    #
+    # The alternative syntax is: nix-tools._raw.cardano-chain.components.$comp
+    # if you want to build only a single component.
```
michaelpj: The first use case seems much less common than the second. Usually if I'm building something I want to build a specific thing, not "all libs in this project".

nix-tools-release.nix
```diff
@@ -0,0 +1,68 @@
...
+
+  mapped-pkgs-all
+    = lib.recursiveUpdate
+        (mapped-pkgs)
+        (lib.mapAttrs (_: (lib.mapAttrs (_: (lib.mapAttrs' (n: v: lib.nameValuePair (lib.systems.examples.mingwW64.config + "-" + n) v)))))
+          mapped-pkgs-mingw32);
+
+in fix (self: other-packages // mapped-pkgs-all
+// {
+#  forceNewEval = pkgs.writeText "forceNewEval" chain.rev;
```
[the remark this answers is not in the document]
angerman: Yes, this still needs to be fixed. I've yet to figure out how to do this nicely.

nix-tools-default.nix
angerman: The origin came primarily for use cases with `exes`, where you want a single derivation and get all the executables instead of having to write each single executable component to get the relevant component. Behind the scenes it's just a joinSymlink (or alternative logic on windows).

nix-tools-release.nix
```diff
@@ -0,0 +1,68 @@
...
+  required = pkgs.lib.hydraJob (pkgs.releaseTools.aggregate {
+    name = required-name;
+    constituents = with self;
+      [ # forceNewEval
+      ] ++ builtins.attrNames other-packages
+        ++ map (pkg: [ nix-tools.exes.${pkg}.x86_64-darwin
+                       nix-tools.exes.${pkg}.x86_64-linux
+                       nix-tools.libs.${pkg}.x86_64-darwin
+                       nix-tools.libs.${pkg}.x86_64-linux ]) packages;
+#               (packages ++ map (pkg: lib.systems.examples.mingwW64.config + "-" + pkg) packages);
```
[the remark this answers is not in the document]
angerman: This should be gone. I think it's much better to be explicit about the jobs you want in your `required` set without any magic.

nix-tools-release.nix
```diff
@@ -0,0 +1,68 @@
...
+          mapped-pkgs-mingw32);
+
+in fix (self: other-packages // mapped-pkgs-all
+// {
+#  forceNewEval = pkgs.writeText "forceNewEval" chain.rev;
+
+  required = pkgs.lib.hydraJob (pkgs.releaseTools.aggregate {
+    name = required-name;
+    constituents = with self;
+      [ # forceNewEval
```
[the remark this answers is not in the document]
angerman: gone :-)

default.nix
angerman: It should be perfectly fine to use the `default-nix` template and just merge it with custom package definitions. The `release-nix` will still rely on the `default.nix`.

angerman: @michaelpj does the example (See the PR message atop) suite your needs?

michaelpj:
> default.nix example

I'd rather do something like this, does that work?
```
let localLib = import ./lib.nix;
in {
    nix-tools = import localLib.nix-tools.default-nix ./nix/pkgs.nix {};
    # other stuff
    ...
}
```

Similarly for the `release.nix`. Basically, I don't like the inversion of control: if I want to have other non-Haskell stuff I have to pass it *in* to the nix-tools template. I'd much rather have it be a plain attribute set or something that I can just use normally.

nix-tools-default.nix
michaelpj: Sure, I'm just saying that I do a lot of `nix build -f default.nix myPackage`, so I'd like the syntax for that to be nice.

default.nix
michaelpj: Good example, do put them somewhere in the actual repository so they're available for future people!

nix-tools-default.nix
angerman: @michaelpj what is `nix build -f default.nix myPackage` supposed to do? You didn't even specify a component? Do you want *all* libs, execs, tests, benchmarks to be built?

michaelpj: `nix build -f default.nix myPackage.lib` is also fine. I'd like `nix build -f default.nix myPackage` to be like `cabal build`-ing it, so yeah, build all the components.

angerman:
@michaelpj maybe I'm misreading you...

> if I want to have other non-Haskell stuff I have to pass it in to the nix-tools template. I'd much rather have it be a plain attribute set or something that I can just use normally.

```nix
{ pkgs, ... }@args:
localLib.nix-tools.default-nix ./nix/pkgs.nix args // {
    # additional non-haskell packages we want to build on CI.
    byronLedgerSpec = import ./specs/ledger/latex {};
    byronChainSpec = import ./specs/chain/latex {};
    semanticsSpec = import ./specs/semantics/latex {};
}
```
This is equivalent to
```nix
{ pkgs, ... }@args:
{
    inherit (localLib.nix-tools.default-nix ./nix/pkgs.nix args) nix-tools;

    # additional non-haskell packages we want to build on CI.
    byronLedgerSpec = import ./specs/ledger/latex {};
    byronChainSpec = import ./specs/chain/latex {};
    semanticsSpec = import ./specs/semantics/latex {};
}
```

The slightly annoying but important part is that we pass `args` to the nix-tools template.  Otherwise cross compilation would not work as hydra through `release.nix`, which will call `default.nix` with arguments, provides the right arguments and pkgs configuration for cross compilation to work.

michaelpj:
Looks like you're still working on this - let me know if you'd like me to have another look.

jbgi:
@angerman Should we also include packaging of `remote-iserv` and friends with this PR? I think it would make sense.

nix-tools-default.nix
```diff
@@ -0,0 +1,42 @@
...
+    nix-tools = { _raw = nix-tools; }
+      # some shorthands
+      // { libs = mapAttrs (k: v: if   v ? components && v.components ? "library"
+                                  then v.components.library
+                                  else null) nix-tools; }
+      // { exes = mapAttrs (k: v: if   (v ? components && length (attrValues v.components.exes) > 0)
+                                  then if pkgs.stdenv.targetPlatform.isWindows then pkgs.copyJoin else pkgs.symlinkJoin
```
jbgi: This `symlinkJoin` logic appears to mess-up with the ability to reference exes in the release.nix `required-targets`... or how do you do that?

angerman: This should be fixed in 04814db.

angerman:
> @angerman Should we also include packaging of `remote-iserv` and friends with this PR? I think it would make sense.

@jbgi yes! yes, we should. Was added in c101f9f.

default.nix
```diff
@@ -55,6 +55,12 @@ let
     regeneratePackages = commonLib.pkgsDefault.callPackage ./nix-tools-regenerate.nix {
       nix-tools = package;
     };
+    # default and release templates that abstract
+    # over the details for CI.
+    default-nix = import ./nix-tools-default.nix (commonLib // { inherit nix-tools; });
+    release-nix = import ./nix-tools-release.nix (commonLib // { inherit nix-tools; });
+    iohk-module = import ./nix-tools-iohk-module.nix commonLib;
+    iohk-overlay = import ./nix-tools-iohk-overlay.nix commonLib;
```
michaelpj: what are these two?

haskell-hackage-stackage.nix
```diff
@@ -0,0 +1,38 @@
+{ pkgs }:
+let
+  # overriding logic so we can pass -I to nix, and overide the
+  # relevant import.
```
michaelpj: I'm sure this makes sense to you, but it's quite mystifying to a reader.

nix-tools-default.nix
```diff
@@ -0,0 +1,45 @@
+commonLib: # the iohk-nix commonLib, that provides access to the pinned packages
+pkgs-path: # the path to th local pkgs.nix file for nix-tools that imports all
```
michaelpj: the

michaelpj: Suggest putting "haskell" in the name of this variable somewhere to avoid confusing it with the `pkgs` that means nixpkgs.

nix-tools-default.nix
michaelpj: Is this comment still accurate?

nix-tools-iohk-module.nix
```diff
@@ -0,0 +1,69 @@
...
+#      iohk-module
+#    ];
+#
+commonLib:
+# we need nixpkgs as this will be the properly configured one
+# the one that gives us the right host and target platforms.
+{ nixpkgs, th-packages ? [] }:
+{ pkgs, buildModules, config, lib, ... }:
+let
+        withTH = import ./mingw_w64.nix {
```
michaelpj: weird indentation

nix-tools-iohk-module.nix
```diff
@@ -0,0 +1,69 @@
...
+# Example:
+#    modules = [
+#      haskell.ghcHackagePatches.${(stack-pkgs.overlay hackage).compiler.nix-name}
+#      iohk-module
+#    ];
+#
+commonLib:
+# we need nixpkgs as this will be the properly configured one
+# the one that gives us the right host and target platforms.
+{ nixpkgs, th-packages ? [] }:
```
michaelpj: what are th-packages?

nix-tools-iohk-overlay.nix
```diff
@@ -0,0 +1,45 @@
...
+    Win32 = hackage.Win32."2.6.2.0".revisions.default;
+  };
+  ghc86 = hackage: {
+    hsc2hs = hackage.hsc2hs."0.68.4".revisions.default;
+    # stackage beautifully omitts the Win32 pkg
+    Win32 = hackage.Win32."2.6.2.0".revisions.default;
+  };
+in {
+  ghc842 = ghc84;
+  ghc843 = ghc84;
```
michaelpj: This was pretty confusing for a bit until I realised that `ghc84` was an overlay.

nix-tools-release.nix
```diff
@@ -0,0 +1,69 @@
+commonLib: # the iohk-nix commonLib
+{ packages ? []
+, required-name ? "required"
+, required-targets ? (jobsets: [])
+, config ? {}
+, rev ? "abcdef"
+, package-set-path # usually import ./. {}
```
michaelpj: is it a path or the imported thing?

nix-tools-release.nix
```diff
@@ -0,0 +1,69 @@
...
+with (import (commonLib.nixpkgs + "/pkgs/top-level/release-lib.nix") {
+  inherit supportedSystems scrubJobs nixpkgsArgs;
+  packageSet = import package-set-path;
+});
+with pkgs.lib;
+let
+
+  traceId = x: builtins.trace (builtins.deepSeq x x) x;
+
+  packageSet = import package-set-path {};
```
michaelpj: should we not be passing some args here?

nix-tools-release.nix
```diff
@@ -0,0 +1,69 @@
...
+    nix-tools.benchmarks =
+      mapAttrs (_: mapAttrs (_: _: supportedSystems))
+        (filterAttrs (n: v: builtins.elem n packages && v != null) packageSet.nix-tools.benchmarks);
+  };
+
+  mapped-pkgs = mapTestOn (nix-tools-pkgs supportedSystems);
+  # we use builtins.currentSystem here as that will evaluate to whatever the evaluator runs on.
+  # thus someone on macOS will be able to build the .x86_64-darwin cross expressions, while
+  # someone on linux will be able to build the .x86_64-linux ones.  As hydra is running on
+  # linux, this should also only present CI with the .x86_64-linux targets.
```
michaelpj: I don't really understand why the mingw32 packages are different in this way. Are they the only cross ones, perhaps?

nix-tools-release.nix
```diff
@@ -0,0 +1,69 @@
...
+
+  mapped-pkgs-all
+    = lib.recursiveUpdate
+        (mapped-pkgs)
+        (lib.mapAttrs (_: (lib.mapAttrs (_: (lib.mapAttrs' (n: v: lib.nameValuePair (lib.systems.examples.mingwW64.config + "-" + n) v)))))
+          mapped-pkgs-mingw32);
+
+in fix (self: (builtins.removeAttrs packageSet ["nix-tools" "_lib"]) // mapped-pkgs-all
+// {
+  forceNewEval = pkgs.writeText "forceNewEval" rev;
```
michaelpj: I guess this is so you can force re-evaluation by changing the rev?

patches/clock-0.7.2.patch
```diff
@@ -0,0 +1,59 @@
+diff --git a/System/Clock.hsc b/System/Clock.hsc
```
michaelpj: Are we generally planning to have our GHC patches here rather than in our nixpkgs fork? Is there a reason for that?

angerman: These are patches for haskell packages. We are not reusing any of the haskell packages from nixpkgs. The packages we use come straight out of the nix-tools infrastructure. Either via a defined stack snapshot or the corresponding translation of a cabal plan file.

nix-tools-default.nix
angerman: Yes. I think this comment is still accurate.

nix-tools-default.nix
angerman: I think `nix-tools` might even be better than `haskell`, as this is the `pkgs.nix` file as generated by the `nix-tools` tooling.

nix-tools-release.nix
angerman: It's actually the path, and the comment is wrong. Good catch!

nix-tools-release.nix
angerman: Yes, mingw32 is the only cross target we have so far. I've extended the comment.

haskell-hackage-stackage.nix
angerman: I've extended the comment to make this clearer.

default.nix
angerman: I've added comments here as well.

nix-tools-release.nix
angerman: We could try to artificially construct arguments, but as we are only interested in the nix-tools attribute, to generate the proper inputs for the mapTestOn and mapTestOnCross functions, an empty set of arguments should do.

nix-tools-release.nix
angerman: I believe that's the idea, this logic was in the release.nix of cardano-chain, and I think I've also seen it in cardano-sl; thus I've just tried to keep it they way it was.

disassembler:
@angerman good with this, but can we as a follow-up add bors since we're adding CI checks now?

PR #31, opened 2019-01-23
Remove compiler override in haskellPackages

michaelpj:
- `pkgsGen` will already have a compiler - if the user wants to set the
compiler this seems like the obvious way to do it.
- Repeatedly overriding a Haskell package set doesn't work very well.
Package sets generated with `stack2nix` will already have had an
override applied, in which case I'm fairly sure this does nothing.

Better just to drop it.


disassembler:
bors ping

PR #36, opened 2019-02-14
Update usage of haskell.nix and nix-tools

rvl:
Haskell.nix provides nix-tools, hackage, and stackage, since input-output-hk/haskell.nix#47.

This updates iohk-nix accordingly. I have to say, the code was very confusing. For example a variable called `nix-tools` has many meanings.


jbgi:
@rvl could you also remove obsolete bits from the `pins/update-defaults.sh` script?

rvl:
@jbgi Done, thanks. In repos which use this, the `mkNewPkgSet` usage needs to change. Here it is for iohk-monitoring: https://github.com/input-output-hk/iohk-monitoring-framework/compare/devops-1222-haskell-nix-update

PR #37, opened 2019-02-19
bump haskell.nix

angerman:


angerman:
Guess we want to do this after we fixed the sublibs (haskell.nix#63) and figured out what's broken with dockets (haskell.nix#66).

angerman:
obsolete.

PR #38, opened 2019-02-24
Use config.nix with haskell by default

angerman:
We have a lot of patches in our config.nix file and those whould be
auto-applied to our package set.  Hence we want commonLib.pkgs instead
of commonLib.pkgsDefault.

### DIFF
--- a/pins/haskell-nix.json
+++ b/pins/haskell-nix.json
@@ -1,7 +1,7 @@
 {
   "url": "https://github.com/input-output-hk/haskell.nix",
-  "rev": "3584345a9ab001d1867e972a1a20b4406cbffd68",
-  "date": "2019-02-14T09:35:57+08:00",
-  "sha256": "08pzfvspfl5nvn5szy7bv3rbwymjgmbdm1ac571c64fzhrwf5ghw",
+  "rev": "0b2d033eefeaf9339320b27a0b159e80a890fc29",
+  "date": "2019-03-05T08:35:36+10:00",
+  "sha256": "0x594a72kyhs4vd3pj401x131ilwjgm4lz6sns8hfb6l166qywf3",
   "fetchSubmodules": false
 }
--- a/default.nix
+++ b/default.nix
@@ -50,11 +50,7 @@ let
     # stack.yaml files.
     package = haskell.nix-tools;
     # A different haskell infrastructure
-    haskell = commonLib.pkgsDefault.callPackage ./haskell.nix {
-      # We need to pass `pkgs` here, otherwise we loose all
-      # config customizations that are essential.
-      pkgs = commonLib.pkgs;
-    };
+    haskell = commonLib.pkgsDefault.callPackage ./haskell.nix;
     # Script to invoke nix-tools stack-to-nix on a repo.
     regeneratePackages = commonLib.pkgsDefault.callPackage ./nix-tools-regenerate.nix {
       nix-tools = package;
--- a/default.nix
+++ b/default.nix
@@ -51,7 +51,9 @@ let
     package = haskell.nix-tools;
     # A different haskell infrastructure
     haskell = commonLib.pkgsDefault.callPackage ./haskell.nix {
-      pkgs = commonLib.pkgsDefault;
+      # We need to pass `pkgs` here, otherwise we loose all
+      # config customizations that are essential.
+      pkgs = commonLib.pkgs;
     };
     # Script to invoke nix-tools stack-to-nix on a repo.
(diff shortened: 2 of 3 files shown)


jbgi:
I'm testing this patch on iohk-monitoring-framework and now I get this different error:
```
builder for '/nix/store/nr7w3xbbb61rr5z7607pc058yfgk6ghz-mingw-w64-5.0.4-pthreads.drv' failed with exit code 2; last 10 log lines:
  libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I./include -DIN_WINPTHREAD -DWINPTHREAD_DBG=1 -Wall -DWIN32_LEAN_AND_MEAN -g -O2 -c src/barrier.c  -fPIC -DPIC -o src/.libs/libwinpthread_la-barrier.o
  src/barrier.c:23:10: fatal error: windows.h: No such file or directory
   #include <windows.h>
            ^~~~~~~~~~~
  compilation terminated.
  make[2]: *** [Makefile:749: src/libwinpthread_la-barrier.lo] Error 1
  make[2]: Leaving directory '/tmp/nix-build-mingw-w64-5.0.4-pthreads.drv-0/mingw-w64-v5.0.4/mingw-w64-libraries/winpthreads'
  make[1]: *** [Makefile:866: all-recursive] Error 1
  make[1]: Leaving directory '/tmp/nix-build-mingw-w64-5.0.4-pthreads.drv-0/mingw-w64-v5.0.4/mingw-w64-libraries/winpthreads'
  make: *** [Makefile:477: all] Error 2
cannot build derivation '/nix/store/lkx6031grv248jgfga3d4q1bs2azgqbx-exceptions-0.10.0-lib-exceptions.drv': 1 dependencies couldn't be built
```

angerman:
@jbgi This only happens for windows cross compilation targets?

jbgi:
@angerman Yes, only for cross compilation target. 
Tried on
https://github.com/input-output-hk/iohk-monitoring-framework/pull/129

`nix build -f release.nix nix-tools.cexes.iohk-monitoring.example-simple.x86_64-linux` works. 
`nix build -f release.nix nix-tools.cexes.x86_64-pc-mingw32-iohk-monitoring.example-simple.x86_64-linux --show-trace` fails with above error.

angerman:
@jbgi: so it *did* work before, and after bumping iohk-nix (and thus including the following changes), it stopped working

https://github.com/input-output-hk/iohk-nix/compare/fdf62bbe2a97f3bbd7ece6d7d6df391ee2c09a11...f20ab3940d3ec5d5a3d92e1f0544f501316adc43

which in turn also included the following changes to `haskell.nix`:
https://github.com/input-output-hk/haskell.nix/compare/70f01c2029b3020c68ae22eef6542c3e7610807d...3584345a9ab001d1867e972a1a20b4406cbffd68

jbgi:
@angerman yes, exactly.

rvl:
Next problem is that there is no `x86_64-pc-mingw32-haddock` build tool for the cross build.
And there is currently no way to globally disable haddock with `haskell.nix` (the `doHaddock` option is at package or component level).

angerman:
> Next problem is that there is no `x86_64-pc-mingw32-haddock` build tool for the cross build.
> And there is currently no way to globally disable haddock with `haskell.nix` (the `doHaddock` option is at package or component level).

I would say we just disable haddock if `target != host`, it's not pretty but it probably does the right thing. With haddock we don't want to run haddock on the target, Thus we really want the haddock on the build machine. At the same time haddock suffers from flattening issues. Can we instruct haddock to set certain flags? Or will it always inherit them from ghc? If you look at the haddocks of packages with conditional CPP, you end up only seeing the haddock part for the system it was built on.

Thus I would just say haddock for the cross compilation case should be disabled for now. If someone wants the haddocks, he should use the non-cross compilation case to get them.

pins/haskell-nix.json
```diff
@@ -1,7 +1,7 @@
 {
   "url": "https://github.com/input-output-hk/haskell.nix",
-  "rev": "3584345a9ab001d1867e972a1a20b4406cbffd68",
-  "date": "2019-02-14T09:35:57+08:00",
-  "sha256": "08pzfvspfl5nvn5szy7bv3rbwymjgmbdm1ac571c64fzhrwf5ghw",
+  "rev": "c6106e3ac07a9f782791f9de750c15a5f13718a1",
```
jbgi: - [ ] replace with actual commit after merge of https://github.com/input-output-hk/haskell.nix/pull/76

PR #39, opened 2019-02-24
Drop iserv-proxy-cleanup patch from ghc844

angerman:
This one only applies cleanly to 85+

### DIFF
--- a/config.nix
+++ b/config.nix
@@ -34,8 +34,8 @@
          # this might be fixed in 8.6.4 (if a release is cut), or 8.8
          ++ lib.optional (builtins.compareVersions drv.version "8.5" ==  1
                        && builtins.compareVersions drv.version "8.8" == -1) ./patches/ghc/MR148--T16104-GhcPlugins.patch
+         ++ lib.optional (builtins.compareVersions drv.version "8.5" ==  1) ./patches/ghc/iserv-proxy-cleanup.patch
          ++ [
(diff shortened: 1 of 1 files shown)


angerman:
So far only `plutus` has been on 8.4.4, which is where I ran into the issue with the patch. Maybe we need to resurrect some older patch to apply to 8.4.4. I remember that the initial `cardano-sl` version was built with 8.4.4. 🤔 

Regarding the `versionOlder` and alike, I'd rather do that in a different PR, as there are lot more places in the `config.nix` that could use that treatment. I'd like to believe I did have a reason *not* to use those version helper though. Memory fails me.

rvl:
> So far only `plutus` has been on 8.4.4, which is where I ran into the issue with the patch. Maybe we need to resurrect some older patch to apply to 8.4.4. I remember that the initial `cardano-sl` version was built with 8.4.4. 

Looks like the setup in `cardano-sl` (which doesn't use `iohk-nix`) is a bit different (I did `git ls-files | grep iserv`).

Given that there is only `plutus` on 8.4.4, unless someone has a plan to use `iohk-nix` with `cardano-sl` then it should be fine.

> Regarding the `versionOlder` and alike, I'd rather do that in a different PR, as there are lot more places in the `config.nix` that could use that treatment. I'd like to believe I did have a reason _not_ to use those version helper though. Memory fails me.

Yep, a separate PR would be fine.

PR #41, opened 2019-03-04
Add commonLib.cache-s3 and stack-hpc-coveralls

rvl:
This is for Buildkite stack builds.

    nix-build -A cache-s3
    nix-build -A stack-hpc-coveralls


rvl:
> Ack. Looks good. Just out of curiosity why don't we use nix-tools here?

Thanks. I didn't use nix-tools because these I just copied & pasted these package expressions from client repos. Also at this stage, there is slightly more boilerplate for haskell.nix.

PR #53, opened 2019-03-21
Adapt stack-to-nix flags to new synthax.

jbgi:


docs/nix-toolification.org
```diff
@@ -243,7 +243,7 @@ nix build -f default.nix _iohk.nix-tools.regeneratePackages --out-link regen-pkg
 This will install ~nix-tools~ and call
 
 #+BEGIN_SRC
-stack-to-nix -o nix stack.yaml > nix/.stack-pkgs.nix~.
+stack-to-nix --output nix --stack-yaml stack.yaml > nix/.stack-pkgs.nix~.
```
angerman:
```suggestion
stack-to-nix --output nix --stack-yaml stack.yaml
```

nix-tools-regenerate.nix
```diff
@@ -32,7 +32,7 @@ in
     }
     trap cleanup EXIT
 
-    stack-to-nix -o nix stack.yaml > "$dest.new"
+    stack-to-nix --output nix > "$dest.new"
```
angerman:
```suggestion
    stack-to-nix --output nix
```

angerman: I don't think `stack-to-nix` generates anything on STDOUT anymore.

But then the two `mv` and `echo` lines below make no sens anymore either.

angerman:
@jbgi I believe the whole `dest` and `dest.new` logic in ` nix-tools-regenerate.nix` doesn't make much sense now anymore? Do you agree?

jbgi:
> @jbgi I believe the whole `dest` and `dest.new` logic in ` nix-tools-regenerate.nix` doesn't make much sense now anymore? Do you agree?

Yes. I am working on a fix right now.

PR #61, opened 2019-04-21
Add more "gotcha" tips

rvl:
Here is some documentation of things I have run into when using iohk-nix/Haskell.nix.

I could move some of this into the Haskell.nix docs some time, but in the meantime it's captured here.

### DIFF
--- a/gotchas/gotcha-2.md
+++ b/gotchas/gotcha-2.md
@@ -1,0 +1,91 @@
+## Haskell.nix Dependency Build Problems
+
+Sometimes the dependencies of a project fail to build and you will
+need to add overrides to the Haskell.nix config to fix it.
+
+This file lists build problems that you may encounter and their solutions.
+
+### `No attribute: hsc2hs`
+
+This is a bug in Haskell.nix which you will see if you have a
+dependency which uses `hsc2hs` as a build tool.
+
+Work around it by adding `hsc2hs` to the `extra-deps` of `stack.yaml`.
+
+### Errors about `Divisible` when compiling `contravariant`
+
+```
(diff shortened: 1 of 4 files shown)


gotchas/gotcha-2.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+This means that the resolver specified in `stack.yaml` was not found
+in `stackage.nix`. The snapshot is probably more recent than your
+version of `stackage.nix`.
+
+The repos `stackage.nix` and `hackage.nix` are updated daily, along
+with their JSON pins in `haskell.nix`.
+
+The best course of action is usually to update your `haskell.nix`
```
angerman: should we provide an example here?

gotchas/gotcha-2.md
```diff
@@ -0,0 +1,91 @@
...
+collection that you are using. Nixpkgs will only keep one compiler of
+each major version.
+
+You may need to use a different Nixpkgs channel.
+
+If using `stack-to-nix`, you could choose a different Stackage snapshot.
+
+If using `plan-to-nix`, you could regenerate your plan and nix with a
+compiler version that exists in your Nixpkgs.
+
```
angerman: If you use `iohk-nix`/`input-output-hk/nixpkgs`, bumping that might help as well.

domenkozar:
Missing gotcha: `pq` dependency missing -> should be added to https://github.com/input-output-hk/haskell.nix/blob/master/lib/system-nixpkgs-map.nix

angerman:
> Missing gotcha: `pq` dependency missing -> should be added to https://github.com/input-output-hk/haskell.nix/blob/master/lib/system-nixpkgs-map.nix

Can you open a PR with the necessary change against `haskell.nix`?

PR #63, opened 2019-05-06
Add linux_cross.nix with iserv logic for iserv-under-qemu

eamsden:


eamsden:
Answering my question above: apparently it does because I get the same issue with recursion even without any iserv stuff coming into play. (i.e. when not doing a cross-compile)

eamsden:
@angerman I figured out the one infinite recursion bug (I wasn't using mkIf when I should have been) but now I'm hitting what appears to be the infinite recursion issue with iserv. I've traced the transitive dependencies as far as I can.

One hint: I can't find anything that depends on "process" and yet it pops up in the --show-trace for the error: https://gist.github.com/eamsden/18af91acaacfab52e9c6237934ad23c3

nix-tools-iohk-module.nix
```diff
@@ -26,69 +27,85 @@ let
...
-in {
-  packages = {
-    # This needs true, otherwise we miss most of the interesting
-    # modules.
-    ghci.flags.ghci = true;
-    # I hope we can apply this globally.
-    ghc.flags.ghci = true;
+  windowsArgs = withTHArgs // { extra-test-libs = [ pkgs.rocksdb pkgs.openssl.bin pkgs.libffi ]; };
+  linuxArgs = withTHArgs // { extra-test-libs = []; };
+in lib.mkMerge [
```
angerman: How is `mkMerge` different from `//`.  I think it's worth adding a comment here, such that the next person don't make the same mistake?

nix-tools-iohk-module.nix
```diff
@@ -13,10 +13,11 @@
 commonLib:
 { pkgs, buildModules, config, lib, ... }:
 let
-  withTH = import ./mingw_w64.nix {
-    inherit (pkgs.stdenv) hostPlatform;
+  withTHArgs ={
```
angerman:
```suggestion
  withTHArgs = {
```

nix-tools-iohk-module.nix
```diff
@@ -26,69 +27,85 @@ let
...
+      libiserv.components.library.doExactConfig = true;
+      # same for iserv-proxy
+      iserv-proxy.components.exes.iserv-proxy.doExactConfig = true;
+      remote-iserv.components.exes.remote-iserv.doExactConfig = true;
+      remote-iserv.postInstall = with pkgs.stdenv; lib.optionalString hostPlatform.isWindows ''
+        cp ${pkgs.libffi}/bin/*.dll $out/bin/
+      '';
 
-    # clock hasn't had a release since 2016(!) that is for three(3) years
-    # now.
```
angerman: This reminds me, this comment isn't correct anymore. As https://github.com/corsis/clock/issues/55 has been closed a week ago with the release of clock-0.8

linux_cross.nix
```diff
@@ -0,0 +1,69 @@
+let
+  # Here we try to figure out which qemu to use based on the host platform.
+  # This guess can be overriden by passing qemuSuffix
+  qemuByHostPlatform = hostPlatform:
```
matthewbauer: There is a qemuArch param if you using a new enough nixpkgs. Just `hostPlatform.qemuArch` should work.

linux_cross.nix
```diff
@@ -0,0 +1,69 @@
...
+    then "aarch64"
+    else abort "Don't know which QEMU to use for hostPlatform ${hostPlatform.config}. Please provide qemuSuffix";
+in
+{ stdenv
+, lib
+, writeScriptBin
+, qemu
+, qemuSuffix ? (qemuByHostPlatform hostPlatform)
+, iserv-proxy
+, remote-iserv
```
matthewbauer: I'm not sure if you even need remote-iserv here? You should be able to just wrap `${qemu}/bin/qemu-${qemuSuffix} ${ghc}/bin/ghc-iserv``` and put it in ```-pgmi```. Haven't tried but might make some of this logic simpler.

linux_cross.nix
```diff
@@ -0,0 +1,69 @@
...
+  # we want this to hold only for arm (32 and 64bit) for now.
+  isLinuxCross = buildPlatform != hostPlatform && hostPlatform.isLinux && (hostPlatform.isAarch32 || hostPlatform.isAarch64);
+  qemuIservWrapper = writeScriptBin "iserv-wrapper" ''
+    #!${stdenv.shell}
+    set -euo pipefail
+    # Unset configure flags as configure should have run already
+    unset configureFlags
+    PORT=$((5000 + $RANDOM % 5000))
+    (>&2 echo "---> Starting remote-iserv on port $PORT")
+    ${qemu}/bin/qemu-${qemuSuffix} ${remote-iserv}/bin/remote-iserv tmp $PORT &
```
matthewbauer: You can also use ```${hostPlatform.emulator pkgs}``` to get a qemu-like emulator.

linux_cross.nix
matthewbauer: Perhaps @angerman knows if there is some way around the complicated iserv logic? It doesn't look like you can just put ghci in ```-pgmi```

angerman: @matthewbauer, you’ll just need to wrap off into a shell script. `-pgmi` expects a single callable. `writeShellScriptBin` with
```
qemu-arm iserv $@
```

Is probably enough. Brevity due to typing this on a phone.

the benefit that iserv-proxy provides is that you can get a better glimpse at what’s going on if at the message level if you pass `-v`.

linux_cross.nix
angerman: @matthewbauer, what’s a qemu-like emulator?

eamsden: @matthewbauer do you mean that `hostPlatform.emulator` will pick out the appropriate emulator for the host platform from the given package set?

eamsden: (A bit of playing around with the repl shows that this is indeed the case, and for the case hostPlatform == buildPlatform, it just gives back a path to a bash shell)

angerman: @eamsden the only thing I'm concerned about here is that `${hostPlatform.emulator pkgs}` is a very recent addition :-/ It won't work with 18.09 as far as I can see. I'd prefer not to be too aggressive with requiring newer nixpkgs channels.

linux_cross.nix
angerman: This one currently misses `aarch64`, I think this would be great to use down the line.  For now I guess we have to handroll this (or just pull it from upstream nixpkgs) :-/

linux_cross.nix
eamsden: angerman: ok well this is not a priority right now anyway. I'd like to see a successful Aarch32 build before I start making things pretty :)

eamsden:
@angerman this works on armv7l now. I'd like to clean it up a bit, try it with just iserv-under-qemu again now that the linker issues are fixed, and pass an override to qemu so it only builds the emulator we need and not all the GTK and Xen stuff.

angerman:
This is part of https://github.com/input-output-hk/haskell.nix

PR #65, opened 2019-05-13
Provide `hackageIndexState` parameter.

angerman:


angerman:
@hamishmack I think this is good to go if we drop d86e66e. Would you mind force-pushing it to `hamishmack:hkm/setup-depends` without that commit?

hamishmack:
Done.  But keep in mind it will probably fail if hackageIndexState is not supported in the haskell.nix used.

angerman:
> Done. But keep in mind it will probably fail if hackageIndexState is not supported in the haskell.nix used.

Right. I'll get around to `haskell.nix` as well :-)

PR #66, opened 2019-05-15
Cleanup the config.nix

angerman:
This tries to cleanup all our ghc patches, and adds the relevant
upstream information.  We also add one more patch relevant for
32bit cross compilation (ghcjs).

### DIFF
--- a/config.nix
+++ b/config.nix
@@ -43,6 +43,7 @@
          ++ lib.optional (versionAtLeast "8.2")                           ./patches/ghc/MR545--ghc-pkg-databases.patch                        # https://gitlab.haskell.org/ghc/ghc/merge_requests/545  -- merged; ghc-8.8.1
          ++ lib.optional (versionAtLeast "8.6")                           ./patches/ghc/outputtable-assert-8.6.patch
          ++ lib.optional (versionAtLeast "8.6")                           ./patches/ghc/mistuke-ghc-err_clean_up_error_handler-8ab1a89af89848f1713e6849f189de66c0ed7898.diff # this is part of Phyx- revamped io-manager.
+         ++ lib.optional (versionAtLeast "8.6.4")                         ./patches/ghc/ghc-8.6.4-reenable-th-qq-in-stage1.patch
          ++ [
           ./patches/ghc/ghc-add-keepCAFs-to-rts.patch                         # https://gitlab.haskell.org/ghc/ghc/merge_requests/950  -- open
           ./patches/ghc/lowercase-8.6.patch                                   # https://gitlab.haskell.org/ghc/ghc/merge_requests/949  -- merged; ghc-8.8.1
@@ -61,8 +62,6 @@
          ++ lib.optional (drv.version == "8.6.3")                         ./patches/ghc/ghc-8.6.3-reinstallable-lib-ghc.patch
          ++ lib.optional (drv.version == "8.6.4")                         ./patches/ghc/ghc-8.6.4-reinstallable-lib-ghc.patch
          ++ lib.optional (drv.version == "8.6.5")                         ./patches/ghc/ghc-8.6.5-reinstallable-lib-ghc.patch
-         ++ lib.optional (drv.version == "8.6.4")                         ./patches/ghc/ghc-8.6.4-reenable-th-qq-in-stage1.patch
-         ++ lib.optional (drv.version == "8.6.5")                         ./patches/ghc/ghc-8.6.4-reenable-th-qq-in-stage1.patch
          ++ lib.optional (drv.version == "8.6.4")                         ./patches/ghc/ghc-8.6.4-better-plusSimplCountErrors.patch
          ;
         # Run autoconf again, because an .ac file may have been patched
--- a/config.nix
+++ b/config.nix
@@ -17,11 +17,13 @@
         enableShared = ps.stdenv.targetPlatform == ps.stdenv.hostPlatform;
         enableIntegerSimple = false;
       };
-    ghcDrvOverrides = drv: {
+    ghcDrvOverrides = drv: y{
         dontStrip = true;
-        hardeningDisable = [ "stackprotector" "format" ];
+        hardeningDisable = (drv.hardeningDisable or []) ++ [ "stackprotector" "format" ];
         patches = (drv.patches or [])
-         ++ lib.optional (builtins.compareVersions drv.version "8.6.3" == 0)  ./patches/ghc/T16057--ghci-doa-on-windows.patch
+         # Patches for which we know they have been merged into a public release already
+         ++ lib.optional (builtins.compareVersions drv.version "8.4.3" == 1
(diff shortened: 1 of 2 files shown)


angerman:
This needs a broad set of tests for all our ghc versions. I'll try to see if I can make the `tools` repo do that job.

config.nix
```diff
@@ -90,14 +98,13 @@
...
-      compiler.ghc862 = (ps.haskell.compiler.ghc862.override ghcPkgOverrides).overrideAttrs ghcDrvOverrides;
-      compiler.ghc863 = (ps.haskell.compiler.ghc863.override ghcPkgOverrides).overrideAttrs ghcDrvOverrides;
-      compiler.ghc864 = (ps.haskell.compiler.ghc864.override ghcPkgOverrides).overrideAttrs ghcDrvOverrides;
-    };
+   haskell = lib.recursiveUpdate ps.haskell {
+       compiler = lib.mapAttrs (_name: compiler: (compiler.override ghcPkgOverrides).overrideAttrs ghcDrvOverrides)
+         # these patches (ghcPkgOverrides and ghcDrvOverrides) only apply to vanilla source ghcs. Not ghcjs or binary distributions.
+         # we also ignore ghc82. And are only concerned with ghc84+
+         (lib.filterAttrs
+           (n: _value: !(ps.stdenv.targetPlatform.isGhcjs or false) # we want to apply this only to non-ghcjs ones. As we do some ghc <- ghcjs mapping for ghcjs.
```
rvl: Maybe use a let binding for this filter predicate so that it can be explained better. like `needsPatches = name: ...`

config.nix
```diff
@@ -19,44 +19,52 @@
...
-         # this might be fixed in 8.6.4 (if a release is cut), or 8.8
          ++ lib.optional (builtins.compareVersions drv.version "8.5" ==  1
                        && builtins.compareVersions drv.version "8.6.4" == -1) ./patches/ghc/MR148--T16104-GhcPlugins.patch
-         ++ lib.optional (builtins.compareVersions drv.version "8.5" ==  1)   ./patches/ghc/iserv-proxy-cleanup.patch
-         ++ lib.optional (builtins.compareVersions drv.version "8.2" ==  1)   ./patches/ghc/MR545--ghc-pkg-databases.patch
          ++ lib.optional (builtins.compareVersions drv.version "8.6.4" == -1) ./patches/ghc/MR95--ghc-pkg-deadlock-fix.patch
+
+         # Patches for which we only know a lower bound.
+         ++ lib.optional (builtins.compareVersions drv.version "8.5" ==  1)
+          ./patches/ghc/iserv-proxy-cleanup.patch                             # https://gitlab.haskell.org/ghc/ghc/merge_requests/250  -- merged; ghc-8.8.1
```
michaelpj: So why not put the 8.8.1 upper bound?

michaelpj: And others below.

config.nix
```diff
@@ -19,44 +19,52 @@
...
-          ./patches/ghc/ghc-8.4.3-Cabal2201-allow-test-wrapper.patch
-          ./patches/ghc/ghc-8.4.3-Cabal2201-response-file-support.patch
-          ./patches/ghc/ghc-8.6-Cabal-fix-datadir.patch
-          ./patches/ghc/MR196--ghc-pkg-shut-up.patch
-         ];
-        postPatch = (drv.postPath or "") + ''
-        autoreconf
-        '';
+         ;
+        postPatch = (drv.postPatch or "") + "\n" + "autoreconf";
```
michaelpj: Can we get a comment about why we need to autoreconf? Shouldn't the configure phase happen after the patch phase?

config.nix
michaelpj: Yeah, this nested map/filter thing is a bit hard to read, some judicious factoring out of functions would help.

config.nix
angerman: Right, but autoreconf, will regenerate the configure script if the .ac files were changed.

michaelpj: So the normal GHC patch phase doesn't autoreconf? I guess if it doesn't do that we have to do it ourselves.

michaelpj: er, configure phase

michaelpj: Or we could put `autoreconfHook` in `nativeBuildInputs`, perhaps.

michaelpj: It looks like it's already there :thinking:

angerman: The underlying reason, and I should and will add a comment is that we may patch the .ac files.  So it's more a fallout from our gracious patching rather than something standard.

angerman: TIL about `autoreconfHook`, I don't see it in the `nativeBuildInputs` of ghc expressions though.

michaelpj: Intriguing: in nixpkgs it appears in the [8.2.2 derivation](https://github.com/input-output-hk/nixpkgs/blob/iohk-18.09/pkgs/development/compilers/ghc/8.2.2.nix#L216) but not the later ones. I have no idea what's gong on here. We could maybe use it but the existing thing is fine.

angerman: Maybe they used to apply a patch that required the autoreconfHook, and when that patch wasn’t required anymore they dropped it.

rvl:
I have rebased this on master and fixed the merge conflict.
I also made some changes so that it's more understandable.
Could you please check that the version ranges are correct?

michaelpj:
This looks good, the only thing I would suggest is putting in upper bounds for the things we know will be in 8.8.1.

michaelpj:
(Can I tell bors to merge things or do I need write access?)

rvl:
I'm not really sure what's going to be in 8.8.1. @angerman will be able to confirm next week.
Bors won't listen to you if you don't have write access, but Plutus team do have write access to this repo.

michaelpj:
Fair enough, let's add the upper bounds later.

config.nix
```diff
@@ -17,48 +17,55 @@
...
+         # Patches for which we know they have been merged into a public release already
+         ++ lib.optional (versionAtLeast "8.4.4" && versionOlder "8.6")   ./patches/ghc/ghc-8.4.4-reinstallable-lib-ghc.patch
+         ++ lib.optional (versionOlder "8.6")                             ./patches/ghc/move-iserv-8.4.2.patch
+         ++ lib.optional (versionOlder "8.6")                             ./patches/ghc/hsc2hs-8.4.2.patch
+         ++ lib.optional (versionOlder "8.6")                             ./patches/ghc/various-8.4.2.patch
+         ++ lib.optional (versionOlder "8.6")                             ./patches/ghc/lowercase-8.4.2.patch
+         ++ lib.optional (versionOlder "8.6")                             ./patches/ghc/cabal-exe-ext-8.4.2.patch
+         ++ lib.optional (versionOlder "8.6")                             ./patches/ghc/ghc-8.4.3-Cabal2201-SMP-test-fix.patch
+         ++ lib.optional (versionOlder "8.6")                             ./patches/ghc/outputtable-assert-8.4.patch
+         ++ lib.optional (versionAtLeast "8.6" && versionOlder "8.6.5")   ./patches/ghc/MR148--T16104-GhcPlugins.patch
```
hamishmack: I think this should have been 8.6.4 see #106 for fix

Issue #72, opened 2019-05-21
Make nixpkgs/haskell.nix overridable via an argument to default.nix?


michaelpj:
At the moment if you want to use a newer/different nixpkgs or haskell.nix you have to either:
- Update the pin in `iohk-nix`, wait for that to be merged, update the pin in your repo.
   - Slow.
   - Can be blocked if other users can't handle the upgrade to the default pin.
- Use the `NIX_PATH` override.
    - Not so easy to encode programmatically - have to modify `nix` invocations, not `.nix` files.

I propose we add a third way: let them be passed as arguments to `default.nix`, and only load them from the pin files if they aren't provided that way.

I think this is complementary to the `NIX_PATH` method, depending what you want. As a bonus, we could get rid of the horrible `application` thing in favour of applications just passing their own specific pin.

angerman:
For nixpkgs this is already possible:

https://github.com/input-output-hk/iohk-nix/blob/c01bbd4e0c55a4101e2078d24046e0b2a731f871/default.nix#L8

For `haskell.nix`, @hamishmack experimented with the same approach.

michaelpj:
Why not have them passed as imported things rather than json specs?

Basically the current way is very IOC - even if you have a `nixpkgs`/`haskell.nix` already, you can't just use it, you have to pass it through `iohk-nix` and use the on it creates.

angerman:
@michaelpj why would you have `nixpkgs` or `haskell.nix` already and not get them through `iohk-nix`? This would mean that any `config` or other applications that `iohk-nix` does over them would result in two different trees. That seems rather intuitive.
I'm tending towards "you are holding it wrong" if you try to pass haskell or nixpkgs in. `iohk-nix` should be the root object you build and then pick nixpkgs, ... from there.

That being said. I'd be happy to convinced otherwise.

michaelpj:
Overall I think it would be nice if `iohk-nix` was as library-like as possible. At the moment it's much more framework-like - you have to let it set stuff up for you. I don't want it to be the root object, I might have other stuff going on that you don't know about!

As an example of how we could do things the other way around, we could have `iohk-nix` looking something like:
```
{
    defaultNixpkgsFn = ... # local pin
    iohkCallNixpkgs = nixpkgs: { config, system, overlays }: nixpkgs { config // iohkConfig, system, overlays ++ iokhOverlays };
   defaultNixpkgs = args :  iohkCallNixpkgs defaultNixpkgsFn args
[...]

PR #74, opened 2019-05-21
Haskell nix override

hamishmack:
I think this should solve #72

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -6,6 +6,7 @@
 , application ? ""
 # Override nixpkgs-src.json to a file in your repo
 , nixpkgsJsonOverride ? ""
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -52,7 +54,7 @@ let
     # stack.yaml files.
     package = (haskell { pkgs = commonLib.pkgsDefault; }).nix-tools;
     # A different haskell infrastructure
-    haskell = import ./haskell.nix;
+    haskell = (import ./haskell.nix) { inherit haskellNixJsonOverride; };
```
angerman: I don't think we strictly need these parens.

PR #77, opened 2019-05-29
Add nixpkgsOverlays arg to modify nixpkgs

hamishmack:
This is a handy way to change parts of nixpkgs without forking the repo

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -6,6 +6,8 @@
 , application ? ""
 # Override nixpkgs-src.json to a file in your repo
 , nixpkgsJsonOverride ? ""
+# Modify nixpkgs with overlays
+, nixpkgsOverlays ? []
 # Override haskell-nix.json to a file in your repo
 , haskellNixJsonOverride ? ""
 }:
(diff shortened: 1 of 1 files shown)


angerman:
@michaelpj I hope this helps you a bit as well.  I don't think we want to allow passing in instantiated `nixpkgs`. The reason is that that would break any form of templating we could do that needs to parameterize over the `nixpkgs` args. That is any form of cross compilation so far needs to build `nixpkgs` with `crossSystem` set properly. If we pass in `nixpkgs` we can't do this anymore.

michaelpj:
> That is any form of cross compilation so far needs to build nixpkgs with crossSystem set properly. If we pass in nixpkgs we can't do this anymore.

I don't buy this. It means that the user needs to pass the same `crossSystem` when they import `iohk-nix` as when they import their `nixpkgs`, but that seems okay. Especially if they're just inheriting all the args.

Also, I think my suggestion in https://github.com/input-output-hk/iohk-nix/issues/72 avoids this by having the user instantiate their nixpkgs using a function provided by iohk-nix, in case we need to do anything clever.

michaelpj:
As a concrete suggestion here, it seems like iohk-nix essentially accepts all the nixpkgs args. Why not have a `nixpkgsArgs` attrset argument that's passed through as is? Then a) it's clear what's happening and b) we avoid gradually adding all the arguments over time.

PR #84, opened 2019-06-05
add the check-hydra.sh script

cleverca22:


ci/check-hydra.nix
```diff
@@ -0,0 +1,15 @@
+{ substituteAll
+, jq
+, hydra
+, lib
+}:
+let
+  check-hydra = substituteAll {
+    src = ./check-hydra.sh;
+    tools = lib.makeBinPath [ jq hydra ];
```
rvl: Technically, it also needs coreutils and time on the PATH.

PR #113, opened 2019-07-03
Use niv to handle pinned dependencies.

jbgi:
(using this fork for niv: https://github.com/input-output-hk/niv/)


rvl:
Good stuff. This will probably improve things.

Could we have the following things documented with this PR:
* A summary of what each of the pins is for.
* How to update niv pins -- either all of them together, or just one. 
* How to update the iohk-nix version in a downstream project
* How to override a single source in a downstream project
* What is the policy for pins - as an example: [the policy for haskell.nix](https://input-output-hk.github.io/haskell.nix/dev/nixpkgs-pin/)

README.md
```diff
@@ -24,20 +24,7 @@ See [Starting a new project](./docs/start.md).
...
-or run the following command:
-
-```
-$ nix-prefetch-git https://github.com/input-output-hk/iohk-nix | tee ./nix/iohk-nix-src.json
-```
-
-(You may need to
-[install](https://nixos.org/nix/manual/#operation-install) the
-`nix-prefetch-git` package beforehand.)
+niv update iohk-nix
```
rvl: How will downstream users get niv?
It has to be an iohk fork of niv, right?

rvl:
@jbgi and @disassembler can you please work together on this?

Get it documented and tested on the reference project (skeleton), _then_ let downstream projects use it (input-output-hk/cardano-wallet#813). Rather than expose them to confusion and churn.

Another thing is that with niv, iohk-nix might work better as an overlay or set of overlays. But this change could go in a separate PR.

disassembler:
we're merging this as-is. Attempted to get skeleton working as good as I could. But the stack rebuild job is impossible to debug. This probably is not a major issue as all the haskell logic is going to be ripped out of this repository in the near future anyways. It just says could not find file default.nix. I did confirm that bumping iohk-nix on cardano-node doesn't break builds in those repositories and shows the deprecation warning.

Issue #117, opened 2019-07-03
Always give a custom error when a non-statically known attribute is accessed


michaelpj:
Many of the gotchas in [gotcha 2](https://github.com/input-output-hk/iohk-nix/blob/master/gotchas/gotcha-2.md) are signalled by "attribute missing" errors. We could significantly improve the UX there (and hopefully not need separate documentation!) by throwing a custom error in those places.

For example, consider the stackage resolver. At the moment if you try and use a resolver that isn't in the current `stackage.nix`, you get an unhelpful error, but we can easily change that:
```diff
       let
         # The Stackage release referenced in the stack config
-        pkg-def = stackage.${stack-pkgs.resolver};
+        pkg-def = stackage.${stack-pkgs.resolver} or throw 
+          "This version of stackage.nix does not know about the Stackage resolver ${stack-pkgs.resolver}, you may need to update haskell.nix to one that includes a newer stackage.nix.";
         # The compiler referenced in the stack config
         compiler = (stack-pkgs.extras hackage).compiler or (pkg-def hackage).compiler;
         patchesModule = ghcHackagePatches.${compiler.nix-name} or {};
```
(I'll open a PR for this one if people like the idea.)

I think there's a pattern here: any time we access an attribute that we don't know *statically* will be present, we should do `or throw <msg>`. Otherwise we're doing the equivalent of `fromJust $ lookup name attrset`.

At least I think we could do this for some of the common cases in the gotchas. And if a user reports an "attribute missing" error and it's not them screwing up, then we should probably add a custom error.

jbgi:
great idea!

michaelpj:
https://github.com/input-output-hk/haskell.nix/pull/194

michaelpj:
I think there's also a bunch of places in the generated nix from `nix-tools` that could do with something like this, but I'm not sure where.

disassembler:
closing in favor of https://github.com/input-output-hk/haskell.nix/pull/194 as this would need implemented in haskell.nix/stackage.nix

michaelpj:
Well, this was supposed to cover the general problem. We do this in a bunch of places (probably in `iohk-nix` too. And that PR doesn't fix the entire problem, so it would be nice to have a tracking issue too.

rvl:
There is also an iohk-nix side to this. If the user updates the LTS version in a cardano project and gets a message "update your haskell.nix", what should they do. They should actually update iohk-nix.

[...]

PR #119, opened 2019-07-04
Bump haskell.nix and update docs

rvl:
Related issue: #117


rvl:
Thanks. I don't think there is really enough information about how to add a new a new GHC version. Maybe that could be moved to a separate section and expanded upon.

bors r+

PR #126, opened 2019-07-09
disable clock-patch

angerman:
This is somehow needed for input-output-hk/cardano-sl#4183

I still don't understand why.


angerman:
This should absolutely not be merged. It will break almost any other package we build. I'm still baffled as to why cardano-sl needs this to be disabled. 😕

PR #129, opened 2019-07-15
Allow control over default-nix for IFD on hydra

hamishmack:
Using callCabalPlanToNix or callStackToNix often results in
timeouts and other eval time errors on hydra that are very difficult
to debug.

We need to be able to ask hydra to build only the `.nix`
files, but not `import` them.

Currently `default-nix` forces the plan to be imported by enumerating
the components it produces.

This change:

* Allows nix-tools-pkgs to be a function to make it possible
  parameterise.
* Allows nix-tools-pkgs to return hsPkgs separately so a the nix
  files can be returned independently.
* Adds nix-tools-raw to get access to nix-tools._raw bypassing the
  enumeration of components.

This change carries the small risk of breaking code that already
includes an `hsPkgs` attribute.

### DIFF
--- a/nix-tools-default.nix
+++ b/nix-tools-default.nix
@@ -19,9 +19,9 @@ let  nix-tools = (if builtins.isFunction nix-tools-pkgs
   inherit pkgs;
   # the iohk-module contains cross compilation specific patches
   inherit (commonLib.nix-tools) iohk-module iohk-extras;
+  };
   # Allow hsPkgs to be returned separately (useful when using
   # callCabalPlanToNix and callStackToNix on Hydra)
-  };
   hsPkgs = nix-tools.hsPkgs or nix-tools;
 in {
     _lib = commonLib;
--- a/nix-tools-default.nix
+++ b/nix-tools-default.nix
@@ -53,7 +53,7 @@ in {
                                   else null) hsPkgs; }
       // { tests = mapAttrs (k: v: if v ? components && length (attrValues v.components.tests) > 0
                                    then v.components.tests
-                                   else null) nix-tools; }
+                                   else null) hsPkgs; }
       // { benchmarks = mapAttrs (k: v: if v ? components && length (attrValues v.components.benchmarks) > 0
                                    then v.components.benchmarks
                                    else null) hsPkgs; }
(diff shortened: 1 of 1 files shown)


nix-tools-default.nix
```diff
@@ -30,23 +39,23 @@ in {
     nix-tools = { _raw = nix-tools; }
```
angerman: Do we still need this, if we have nix-tools-raw?

nix-tools-default.nix
```diff
@@ -17,10 +19,17 @@ let  nix-tools = import nix-tools-pkgs-path {
...
+  };
+  # Allow hsPkgs to be returned separately (useful when using
+  # callCabalPlanToNix and callStackToNix on Hydra)
+  hsPkgs = nix-tools.hsPkgs or nix-tools;
 in {
     _lib = commonLib;
 
+    # Unlike `nix-tools._raw`, accessing `nix-tools-raw` does not force
+    # the components to be enumerated.
+    nix-tools-raw = nix-tools;
```
angerman: shouldn't this be `nix-tools-raw = hsPkgs`?

nix-tools-default.nix
hamishmack: I left it in for backwards compatibility.

nix-tools-default.nix
hamishmack: We need to be able to return something other than `hsPkgs` from `nix-tools-pkgs` and get access to it through `nix-tools-raw`.  In particular the nix file generated by callCabalPlanToNix.  In the case that `nix-tools-pkgs` returns both `plan-nix` and `hsPkgs` we need to be able to access both via `nix-tools-raw.plan-nix` and `nix-tools-raw.hsPkgs`.  That way we can get the `plan-nix` without ever causing it to be imported (by looking at `hsPkgs`).

nix-tools-default.nix
jbgi: note that backward compatibility is already "broken" by removing `nix-tools-pkgs-path`.

jbgi: on that note, please update `changelog.md` so that people can easily make the necessary changes after updating iohk-nix in their project.

PR #130, opened 2019-07-15
Reduce closure size of regenerate script

rvl:
See also: input-output-hk/haskell.nix#178

Tested by running and inspecting the regen script:

    NIX_PATH=iohk_nix=$HOME/iohk/iohk-nix nix-build nix/iohk-common.nix -A nix-tools.regeneratePackages -o regen
    du -scm $(nix-store -qR regen) | sort -n | tail -n25

It makes a difference of about 100MB. If we could kick gcc out of the nix-tools closure, that would make a significant difference.

### DIFF
--- a/nix-tools-regenerate.nix
+++ b/nix-tools-regenerate.nix
@@ -1,11 +1,10 @@
 # A script for generating the nix haskell package set based on stackage,
 # using the common convention for repo layout.
 
-{ lib, stdenv, path, writeScript, nix-tools, coreutils, findutils, gawk
-, nix, nix-prefetch-scripts }:
+{ lib, stdenv, path, writeScript, nix-tools, coreutils, findutils }:
 
 let
-  deps = [ nix-tools coreutils gawk nix nix-prefetch-scripts findutils ];
+  deps = [ nix-tools coreutils findutils ];
 
 in
   writeScript "nix-tools-regenerate" ''


nix-tools-regenerate.nix
```diff
@@ -1,11 +1,10 @@
 # A script for generating the nix haskell package set based on stackage,
 # using the common convention for repo layout.
 
-{ lib, stdenv, path, writeScript, nix-tools, coreutils, findutils, gawk
-, nix, nix-prefetch-scripts }:
+{ lib, stdenv, path, writeScript, nix-tools, coreutils, findutils }:
 
 let
-  deps = [ nix-tools coreutils gawk nix nix-prefetch-scripts findutils ];
```
angerman: I'm not sure we can drop `nix-prefetch-scripts`. I believe `stack-to-nix` will call out to the prefetch commands for source repositories. See

https://github.com/input-output-hk/nix-tools/blob/bb5700d04f0ff2d1a3a7b3278121c1ec67898793/lib/Distribution/Nixpkgs/Fetch.hs#L145

rvl: Yes, we can drop it because the nix-tools programs are [wrapped](https://github.com/input-output-hk/haskell.nix/blob/4c5cf05ab4e433cb6de1adc56f16166542d55a02/nix-tools/default.nix#L50-L54).

angerman: I see. I've added input-output-hk/haskell.nix#208

PR #131, opened 2019-07-16
Improved check-nix-tools script

rvl:
This is the script that's being used in cardano-wallet.

- The script will now detect new files - previously it would miss them and allow the build to break.

- Generate a more convenient patch with "git format-patch".

- The patch has a better filename, which is more convenient to download with the web browser.

- If a SSH key with repo access is installed on the Buildkite agent, it will push changes back to the PR branch.

- Commit signing is not implemented yet. This requires a GPG key deployed on the Buildkite agents.

- In the event of multiple regenerations, subsequent commits are made with [`git commit --fixup`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---fixupltcommitgt). Then it's up to the user whether they would like to squash with [`git rebase --autosquash`](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash).


ci/check-nix-tools.nix
```diff
@@ -14,21 +12,68 @@ writeScript "check-nix-tools.sh" ''
...
+      git format-patch --stdout -1 HEAD > "$patch"
+      buildkite-agent artifact upload "$patch" --job "$BUILDKITE_JOB_ID"
+
+      # Push the changes back to the pull request
+      if [ -n "''${BUILDKITE_PULL_REQUEST_REPO:-}" ]; then
+        sshkey="/run/keys/buildkite-$BUILDKITE_PIPELINE_SLUG-ssh-private"
+        if [ -e $sshkey ]; then
+          echo "Authenticating using SSH with $sshkey"
+          export GIT_SSH_COMMAND="ssh -i $sshkey -F /dev/null"
+          remote=$(echo $BUILDKITE_PULL_REQUEST_REPO | sed -e 's=^[a-z]*://github.com/=git@github.com:=')
```
angerman: what do we expect here other than `https` or `http`?

rvl: `git://` ... thanks Buildkite.

angerman: amazeballs!

rvl:
This will only push updates for Buildkite pull request builds.

If there is a Buildkite branch build without a pull request, it will generate a patch and fail, but not push.

Maybe we want to change that?

angerman:
> This will only push updates for Buildkite pull request builds.
> 
> If there is a Buildkite branch build without a pull request, it will generate a patch and fail, but not push.
> 
> Maybe we want to change that?

I find this to be a rather sensible decision. We wouldn't want it to push to master would we? Fixing up pull requests already feels a bit invasive, but acceptable.  Fixing up arbitrary branches feels wrong (to me).

ci/check-nix-tools.nix
```diff
@@ -14,21 +12,68 @@ writeScript "check-nix-tools.sh" ''
...
+          git push $remote HEAD:$BUILDKITE_BRANCH
```
jbgi: since there is a possibility of a `--fixup` commit, this line must assume a possible force push or rather a `--force-with-lease` push to avoid overriding user commits pushed between initial fetch and the force push.

ci/check-nix-tools.nix
```diff
@@ -14,21 +12,68 @@ writeScript "check-nix-tools.sh" ''
...
+      commit_fixup=$(git log --pretty=oneline --no-decorate origin/$BUILDKITE_PULL_REQUEST_BASE_BRANCH..HEAD | awk "/$commit_message/ { print \$1; }")
+    fi
+
+    # Create the commit
+    export GIT_COMMITTER_NAME="IOHK"
+    export GIT_COMMITTER_EMAIL="devops+nix-tools@iohk.io"
+    export GIT_AUTHOR_NAME="$GIT_COMMITTER_NAME"
+    export GIT_AUTHOR_EMAIL="$GIT_COMMITTER_EMAIL"
+    if [ -n "$commit_fixup" ]; then
+      git commit --no-gpg-sign --fixup "$commit_fixup"
```
jbgi: Since this lead to force pushes, I am not too sure I am in favor of it. This will inevitably breaks some developers workflows. At the minimum I would prefer we only fixup with the last commit but not further ancient commit.

rvl: [`git commit --fixup`](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---fixupltcommitgt) will not overwrite git history. It adds a new commit. Then it's up to the user whether they would like to squash with [`git rebase --autosquash`](https://git-scm.com/docs/git-rebase#Documentation/git-rebase.txt---autosquash). This is quite cool I think.

ci/check-nix-tools.nix
rvl: There is no `--force` here. The push will fail if non-fast-forward. Which will be fine, because Buildkite will re-run this script at the new branch HEAD anyway.

ci/check-nix-tools.nix
jbgi: ah very nice indeed!

rvl:
Hmm bors cancelled for some reason.

bors r+

rvl:
Maybe it didn't like the negative review on this PR.

bors r+

PR #132, opened 2019-07-17
add source-pins to iohk-nix

cleverca22:


angerman:
Out of curiosity, what does this buy us?

rvl:
For the benefit of users and other people who need to maintain the code, please include adequate documentation with your changes.

cleverca22:
@angerman it forces hydra to include the eval-time deps (IFD issues) in the runtime closure of at least one job, so hydra is then forced to upload those to the binary cache

it solves the issue where nix has to build things at eval time, because hydra was not aware of the dependency

PR #133, opened 2019-07-18
Update pins

erikd:
I have no idea what I am doing.


erikd:
Sam suggested this would help.

PR #135, opened 2019-07-18
Define aggregate jobs in nix-tools-release.nix

jbgi:
so that explicitly listings all tests or all exes
 in required jobs is not needed anymore.

Demo: https://github.com/input-output-hk/ouroboros-network/commit/1c659abeeda89f4d6ec39f80fd738d12c25e2617

### DIFF
--- a/nix-tools-release.nix
+++ b/nix-tools-release.nix
@@ -77,7 +77,31 @@ let
         (lib.recursiveUpdate mapped-pkgs-partial extraBuilds)
         mapped-builds-on-supported-systems;
 
-in fix (self: (builtins.removeAttrs packageSet ["nix-tools" "_lib"]) // mapped-pkgs-all
+  aggregate-pkgs = let
+    w64 = lib.systems.examples.mingwW64.config;
+    cS = builtins.currentSystem;
+  in foldr (cType: (lib.recursiveUpdate {
+    # we define aggregate jobs that respectively gather all libs, all exes, all tests and all benchmarks
+    # in the `packages` list.
+    nix-tools."packages-${cType}" = listToAttrs (map (s: nameValuePair s (pkgs.lib.hydraJob (pkgs.releaseTools.aggregate {
+      name = "packages-${cType}.${s}";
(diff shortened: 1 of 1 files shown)


rvl:
I think this change would be fine, except maybe the aggregate derivations will cause slow builds due to copying between build slaves (I'm not sure).

How I did it in cardano-wallet. In [`cardano-wallet/release.nix`](https://github.com/input-output-hk/cardano-wallet/blob/master/release.nix) there is a function
```
collectTests = ds: filter (d: elem d.system testsSupportedSystems) (collect isDerivation ds);
```
The result of this is added to the "required" aggregate job. I think this reduces duplication but still allows control. For example, we usually don't want to run all tests on non-linux builds (for resources reasons).

The cardano-wallet build is centred around a few new utility functions.

1. `isCardanoWallet` - I would like to generalise this and add to nix-tools/haskell.nix.
2. `collectComponents` - This should be added to haskell.nix.
3. `release-lib.nix` - A fork of the NixOS one - we could add this to `iohk-nix`.

angerman:
> The cardano-wallet build is centred around a few new utility functions.
> 
> 1. `isCardanoWallet` - I would like to generalise this and add to nix-tools/haskell.nix.
> 2. `collectComponents` - This should be added to haskell.nix.
> 3. `release-lib.nix` - A fork of the NixOS one - we could add this to `iohk-nix`.

I think we definitely should do 2 and 3. That is add them to haskell.nix, and iohk-nix, that would reduce the boilerplate in cardano-wallet, and allow other projects to easily make use of them as well.

jbgi:
> except maybe the aggregate derivations will cause slow builds due to copying between build slaves (I'm not sure).

good point. I just had a look at [`aggregate` definition](https://github.com/NixOS/nixpkgs/blob/92a047a6c4d46a222e9c323ea85882d0a7a13af8/pkgs/build-support/release/default.nix#L56) and it have `preferLocalBuild = true;` which, I think, means that nesting aggregates should not cause any supplementary network load (vs a single aggregate), since all aggregates should build on the same machine.

PR #148, opened 2019-07-31
add cardano overlay

disassembler:
Adds jormungandr and http bridge cardano-wallet releases to iohk-nix


angerman:
I believe this is stale.

PR #149, opened 2019-08-03
Replace "versionOlder" with "versionLessThan"

hamishmack:
An older version is a version of a package when it was younger.
I mixed this up and I suspect others might too.  I think
`versionLessThan` is harder to confuse.

### DIFF
--- a/config.nix
+++ b/config.nix
@@ -18,25 +18,25 @@
         enableIntegerSimple = false;
       };
     ghcDrvOverrides = drv: let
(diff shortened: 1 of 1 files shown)


rvl:
@hamishmack This makes builds fail with:

```
error: undefined variable 'versionLessThan' at /nix/store/6fldvjyf2ykgwag9swkl1vchjqcqlnz6-source/config.nix:24:30
```

I think you need to `let versionLessThan = versionOlder;` because that's what it's called in `nixpkgs.lib`.

PR #151, opened 2019-08-08
allow override of nixpkgs instead of src json

disassembler:
We're experimenting with switching to niv for version tracking of individual repos in iohk-ops. iohk-nix currently only allows you to override nixpkgs rev using a json file iohk nix expects, but we want to override with a path to the source of nixpkgs (generated by niv).


michaelpj:
OOI why not pass the imported (but not called) nixpkgs function? Is there an advantage to passing the path without importing it?

PR #160, opened 2019-08-23
check-nix-tools: Better error message when check fails

rvl:


deepfire:
I'm properly scared of CI machines being able to push to Github.

PR #166, opened 2019-09-03
Nixpkgs pin 19.03

disassembler:
Updates default pin to 19.03. Bumps all other pins.


michaelpj:
What's holding this up? Are we going to shoot for 19.09?

PR #167, opened 2019-09-04
Add jormungandr-master.

jbgi:


rvl:
I added functions for building custom versions. Caveat is that it probably won't work all the time, depending on the revision used. We will use this in cardano-wallet to pin the jormungandr version.

PR #178, opened 2019-09-30
Add package for latest stylish-haskell from github

rvl:
The stylish-haskell that comes with nixpkgs 18.09 haskellPackages is a little bit old and missing support for some GHC extensions (e.g. DerivingVia).


craigem:
Would it be more graceful to just bump the version in `pkgs/development/haskell-modules/hackage-packages.nix`?

rvl:
@craigem Try it -- good luck. :-) That's a much older LTS than the one stylish-haskell is using.

Anviking:
@rvl is on leave… is it ok to merge this?

PR #195, opened 2019-10-14
Fix the import call for callStackToNix

angerman:
this was introduced in a recent haskell.nix change which tried to align
callCabalProjectToNix and callStackToNix


angerman:
To get this fully working, we'll need a bump of haskell.nix past https://github.com/input-output-hk/haskell.nix/pull/263 as well.

PR #208, opened 2019-10-25
Add cached stack builds to skeleton projects

rvl:
Continuous testing of build code such as iohk-nix and Haskell.nix is necessary and things break when it isn't done.

This adds back the Buildkite jobs for skeleton.

I took the `rebuild.hs` script from cardano-wallet and put it into a Cabal package. A nix function `haskellBuildUtils.stackRebuild { ... }` provides access to this package for project build scripts.

The iohk-skeleton reference project is updated to use the build library.

The stack build script now has caching, which greatly improves CI performance.

@dnadales has offered to do the heavy lifting and make the build library sensible and suitable for use in multiple projects.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,9 +11,9 @@ steps:
     agents:
       system: x86_64-linux
 
-#  - label: ":pipeline: Skeleton project pipelines"
-#    command:
-#      - 'buildkite-agent pipeline upload skeleton/.buildkite/pipeline.yml'
-#      - 'buildkite-agent pipeline upload skeleton/.buildkite/nightly.yml'
-#    agents:
-#      system: x86_64-linux
+  - label: ":pipeline: Skeleton project pipelines"
+    command:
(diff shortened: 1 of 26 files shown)


skeleton/.buildkite/rebuild.hs
```diff
@@ -1,69 +1,155 @@
...
-run cmd args = do
-  printf (s % " " % s % "\n") cmd (T.unwords args)
-  res <- proc cmd args empty
-  case res of
-    ExitSuccess      -> pure ()
-    ExitFailure code -> eprintf
-      ("error: Command exited with code " % d % "!\nContinuing...\n")
-      code
-  pure res
+    (RebuildOpts{..}, cmd) <- parseOpts
```
dnadales: I can imagine all the projects will need to do command line parsing. Shouldn't we put this in a separate library as well?

rvl: Yes perhaps. I wasn't sure whether some projects would need more/different options.

PR #209, opened 2019-10-27
release-lib: make mkRequired build-version work to force evals

disassembler:
This fixes mkRequired to force evals for every revision.


rvl:
@disassembler Why does the `version.json` derivation need to be a top-level job? Is it not enough for it to be a constituent of `required`?

PR #215, opened 2019-10-31
Use overridden nixpkgs in getPkgs

michaelpj:
I think this was just an oversight: it uses the (potentially overridden)
nixpkgs JSON, but not the nixpkgs source if that is passed instead.


michaelpj:
prod

PR #217, opened 2019-11-05
Move build arguments parsing functionality to a separate module.

dnadales:
Edit: this should be ready for review.

---

I started with a draft PR to make sure I don't break anything on CI while moving things from the skeleton project.

My plan is to abstract as much as possible, incorporating the parameters I know other projects will need, so that it is easier to incorporate the `iohk-nix-utils` library in the other repos.

We can make `iohk-nix-utils` more parametric as the need for this arises.


.gitignore
```diff
@@ -1,4 +1,12 @@
...
 \#*\#
 *.*~
-*.bak
\ No newline at end of file
+*.bak
+
+## Editor artifacts and configuration files
+TAGS
+.dir-locals.el
+*.md.temp
```
KtorZ: Slight note but IMO these (as well as any OS-related files) shouldn't part of a project `.gitignore` for these only regard your own setup. 

Instead, I'd recommend using a global `.gitignore` for your shell using the `core.excludesfile` configuration option for git.

**~/.gitconfig**
```gitconfig
[core]
  excludesfile = /home/ktorz/.gitignore # Always ignore these files, in every git project
```

which in my case, points to the following:

**~/.gitignore**
```gitignore
### OSX Bullshit ###
.DS_Store
Thumbs.db
*.orig

### Tags File ###
tags

### Git Worktrees ###
.worktrees

### Vim ###
[._]*.s[a-w][a-z]
[._]s[a-w][a-z]
*.un~
Session.vim
.netrwhist
*~

### Stack ###
.stack-work-* # extra working trees manually created sometimes (e.g. .stack-work-profiling)
```

skeleton/.buildkite/rebuild.hs
```diff
@@ -1,155 +1,56 @@
...
+  doBuild
+    (LibraryName "skeleton")
+    Fast
+    qaLevelDependingOnBorsBranch
+    [TestRun (TestArguments []) []]
+    (CoverallsConfig
+       (CoverallsTokenEnvVar "SKELETON_COVERALLS_REPO_TOKEN")
+       (ExtraShcArgs [])
+       (TixDirectory "lib")
+    )
```
KtorZ: :+1:

utils/lib/Build.hs
```diff
@@ -0,0 +1,177 @@
...
+    CleanupCache ->
+      cleanupCacheStep optDryRun cacheConfig optBuildDirectory
+    PurgeCache ->
+      purgeCacheStep optDryRun cacheConfig optBuildDirectory
+
+newtype LibraryName = LibraryName String deriving Show
+
+data Optimizations = Standard | Fast deriving (Show, Eq)
+
+data QALevel = QuickTest | FullTest deriving (Show, Eq)
```
KtorZ: I think the `QALevel` is perhaps specific to our own project on `cardano-wallet` because we've categorized tests in two categories:

- unit
- integration

For example:  `cardano-wallet-jormungandr:test:unit` and `cardano-wallet-jormungandr:test:integration`. The latter requires extra setup as it needs to connect to an actual running node. 

I am not sure this peculiarity is common to many projects and should be part of the skeleton code. Could be taken away IMO, or rebranded simply as "skip" options being passed down to stack (the stack `--skip`, not the hspec one).

utils/lib/CommonBuild.hs
```diff
@@ -150,16 +154,46 @@ weederStep dryRun = do
...
+-- If coverage information shouldn't be uploaded, or such token is not
+-- available, then the string can be set to "".
+newtype CoverallsTokenEnvVar = CoverallsTokenEnvVar Text deriving (Show)
+
+-- | Extra arguments to be passed to the @shc@ program, which provides
+-- coveralls.io integration to @stack@ based projects.
+newtype ExtraShcArgs = ExtraShcArgs [Text] deriving (Show)
+
+-- | Path to the directory where the coverage files (.tix) can be found.
+newtype TixDirectory = TixDirectory FilePath deriving (Show)
```
KtorZ: Note that this isn't what `tixDir` is used for down there :upside_down_face: 
The tix files can be found in `stack path --local-hpc-root` which is basically what `shc` uses under the hood. 

Here however, we are looking for extra tix files possibly generated by executables built with `--coverage` and ran during the test suite of another package! For example, in the case of `cardano-wallet`, during the integration tests, we actually make calls to the `cardano-wallet` command-line to have automated tests of this command line. This is done via the [process](https://hackage.haskell.org/package/process-1.6.6.0) library and therefore, execution of commands happen in _another process_ that the one executing the test suite. Therefore, `shc` and `stack` and not able to capture this properly. Yet, if the command-line is built using `--coverage` (and this is why this option is passed during the `build` step), it'll generate a corresponding `.tix` file at the root of the project. 

This extra `.tix` tile can then be passed to `shc` when using the `combined` and `custom` options (combined means, all "normally" generated .tix files for the project located in `stack path --local-hpc-root` plus, the extra tix files given as arguments. 

I believe it is possible to figure out what are the roots of each package in the project (in our case, it was simply every folder in `lib/`) from stack or a `cabal.project` or simply from the `stack.yaml` and remove the need for this option altogether :thinking: ?

.gitignore
dnadales: That's a good idea, however I'm worried that this would rely on each developer setting up her own `~/.gitconfig` properly.

KtorZ: Yes. It does :)

PR #238, opened 2019-11-20
Add shelley-staging-short network config.

jbgi:


cardano-lib/shelley-staging-short-config.nix
```diff
@@ -0,0 +1,72 @@
...
+  CriticalCQBootstrap = 0.8888;
+  CriticalCQ = 0.654321;
+  CriticalForkThreshold = 3;
+  FixedTimeCQ = 3600;
+
+  SlotLength = 20000;
+  NetworkConnectionTimeout = 15000;
+  HandshakeTimeout = 30000;
+
+  CA-Organization = "Input Output HK";
```
disassembler: Try removing these. I requested they drop them from being required as they aren't used.

cardano-lib/shelley-staging-short-config.nix
```diff
@@ -0,0 +1,67 @@
+##########################################################
+###############          Testnet           ###############
```
johnalotoski: Suggest changing `Testnet` to `shelley-staging-short` in this header

PR #242, opened 2019-11-22
skeleton: example src fixes

rvl:
These are some fixes that @manveru and @dnadales needed to make for input-output-hk/cardano-ledger-specs#1009.

Note that new versions of Haskell.nix have a
`haskell-nix.haskellLib.cleanGit` filter that provides a `subDir`
filter, which could be used here (thanks @hamishmack). However skeleton is still stuck on an old Haskell.nix version.

It would be useful if the skeleton project were updated to use the latest revision of Haskell.nix, because projects need to upgrade.

### DIFF
--- a/skeleton/nix/pkgs.nix
+++ b/skeleton/nix/pkgs.nix
@@ -33,10 +33,10 @@ let
   pkgSet = haskell.mkStackPkgSet {
     inherit stack-pkgs;
     modules = [
-      # Add source filtering to local packages
+      # Add filtered sources to local packages
       {
         packages.iohk-skeleton.src = src;
-        # packages.another-package = src + /another-package;
+        # packages.another-package.src = src + /another-package;
       }
 
       # Add dependencies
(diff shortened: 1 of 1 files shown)


skeleton/nix/pkgs.nix
```diff
@@ -33,10 +33,10 @@ let
   pkgSet = haskell.mkStackPkgSet {
     inherit stack-pkgs;
     modules = [
-      # Add source filtering to local packages
+      # Add filtered sources to local packages
       {
         packages.iohk-skeleton.src = src;
-        # packages.another-package = src + /another-package;
+        # packages.another-package.src = src + /another-package;
```
dnadales: Just for future reference I guess this is what people needs to do if the project has multiple directories:

https://github.com/input-output-hk/cardano-ledger-specs/pull/1009/commits/25d7ae13599a45da197783f38c132266c1c94d5f#diff-942fcbe81271545077c2ade39eea9a6eR31

I wonder whether `iohk-nix` should provide such `filterSubDir` function.

PR #257, opened 2019-12-02
jormungandr: add 0.8.0-rc4

manveru:


jormungandr-lib/default.nix
```diff
@@ -74,15 +80,12 @@ let
 
   mkConfig = environment:
     let
```
jbgi: empty `let` could be removed.

PR #262, opened 2019-12-03
Convert skeleton to use niv

jbgi:


skeleton/default.nix
```diff
@@ -8,16 +8,16 @@
 { system ? builtins.currentSystem
 , crossSystem ? null
 , config ? {}
-# Import IOHK common nix lib
-, commonLib ? import ./lib.nix
-# Use nixpkgs pin from commonLib
-, pkgs ? commonLib.pkgs
+, sourcesOverride ? {}
+, pkgs ? import ./nix {
```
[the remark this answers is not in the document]
jbgi: Added comments!

I'm a bit ambivalent on `nix/nixpkgs.nix`, it is probably more explicit, but also more verbose and in some projects (like `cardano-ops`) we `import ../nix {}` at numerous places. Also using `nix/default.nix` seems to be the convention used by projects using `niv`, eg.:
 - https://github.com/nmattia/niv/blob/master/nix/default.nix
 - https://github.com/hercules-ci/gitignore.nix/blob/master/nix/default.nix
 - https://github.com/cachix/cachix/blob/master/nix/default.nix

skeleton/nix/default.nix
```diff
@@ -0,0 +1,28 @@
...
+  nixpkgsOverlays = [
+    (_: _: { commonLib = lib // iohkNix; })
+  ];
+
+  # Import IOHK common nix lib, using our sources as override:
+  iohkNix = import sources.iohk-nix {
+    inherit system crossSystem config nixpkgsOverlays;
+    sourcesOverride = sources;
+  };
+  pkgs = iohkNix.pkgs;
```
michaelpj: I still really dislike the weird proxying of the nixpkgs arguments through `iohk-nix`, but I guess that's orthogonal to this PR.

skeleton/nix/sources.json
```diff
@@ -0,0 +1,50 @@
...
+        "homepage": "https://github.com/input-output-hk/niv",
+        "owner": "input-output-hk",
+        "repo": "niv",
+        "rev": "89da7b2e7ae0779fd351618fc74df1b1da5e6214",
+        "sha256": "0jn88ahv6mycxdkzihy1v0dz2110q0d42q4ggyj68l2lyba6y0yy",
+        "type": "tarball",
+        "url": "https://github.com/input-output-hk/niv/archive/89da7b2e7ae0779fd351618fc74df1b1da5e6214.tar.gz",
+        "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
+    },
+    "nixpkgs": {
```
michaelpj: I take it this is just here to make `niv` happy? Wish they would do https://github.com/nmattia/niv/issues/134 already...

jbgi: indeed, otherwise we have to rely on <nixpkgs>.

skeleton/nix/default.nix
jbgi: yeah, would probably be best if iohk-nix expose an ìohk-overlays` and a `iohk-config`. and people use them explicitly.

PR #317, opened 2020-01-31
Update skeleton to cabalProject and overlays, and use nixpkgs19.09 as default

jbgi:


default.nix
```diff
@@ -65,12 +65,13 @@ let
...
       crossSystem' = crossSystem;
     in { args ? {}
+       , overlays ? [ jemallocOverlay ]
        , extraOverlays ? nixpkgsOverlays
        , system ? system'
        , globalConfig ? globalConfig'
        , config ? config'
        , crossSystem ? crossSystem' }: import nixpkgs ({
-          overlays = [ jemallocOverlay ] ++ extraOverlays;
+          overlays = overlays ++ extraOverlays;
```
angerman: Wouldn't we want `overlays ++ [ jemallocOverlay ] ++ extraOverlays`? lest we loose the `jemallocOverlay` once someone passes in the `overlays` arg?

jbgi: yeah, though existing behavior is preserved. only a new capacity to set all overlays is added.

jbgi: ah also, apparently `jemallocOverlay` is only useful for `18.03`, the new jemalloc version that fix the issue is the default since `18.09`.

default.nix
```diff
@@ -16,8 +15,10 @@
 
 let
   defaultSources = import ./nix/sources.nix;
-  pkgsDefault = import (defaultSources.nixpkgs) {};
+  pkgsDefault = import defaultSources.nixpkgs (import defaultSources."haskell.nix");
```
disassembler: I don't like this it essentially means you need haskell.nix builtin the overlay for projects that don't use haskell at all.

default.nix
```diff
@@ -125,12 +119,12 @@ let
   nix-tools = rec {
     # Programs for generating nix haskell package sets from cabal and
     # stack.yaml files.
-    package = nixToolsDeprecation (haskell { pkgs = commonLib.pkgsDefault; }).nix-tools;
+    package = nixToolsDeprecation pkgsDefault.haskell-nix.nix-tools;
     # A different haskell infrastructure
-    haskell = { pkgs }: nixToolsDeprecation import sources.haskell { inherit pkgs; };
+    haskell = _: nixToolsDeprecation (commonLib.getPkg {}).haskell-nix;
     # Script to invoke nix-tools stack-to-nix on a repo.
     regeneratePackages = nixToolsDeprecation commonLib.pkgsDefault.callPackage ./nix-tools-regenerate.nix {
```
disassembler: lets rename this regenerateStackPackages and the deprecation can be removed since the wallet team is utilizing the stack based workflow.

default.nix
```diff
@@ -167,9 +161,7 @@ let
       (import ./overlays/rust)
     ];
     pkgs = import sources.nixpkgs-rust {
```
disassembler: now that we have a sane version of nixpkgs, can we remove nixpkgs-rust?

jbgi:
bors ping

PR #318, opened 2020-02-03
Haskell overhaul

disassembler:
adds helpers for cabalProject. Switches default nixpkgs to upstream.


cabal-project-regenerate/result
```diff
@@ -0,0 +1 @@
+/nix/store/1y97wzhdc9wjwlxs2mbnz96i4zgyybnh-cabal-project-regenerate
```
johnalotoski: result* to gitignore?

PR #328, opened 2020-02-17
Add checkRepoTagsOnMaster: ensure dependencies are on master branches

jbgi:


jbgi:
btw we should probably do a similar script for niv dependencies...

PR #341, opened 2020-02-24
nix-tools-regenerate fix

jbgi:


default.nix
```diff
@@ -121,6 +121,18 @@ let
...
 
+  haskell-nix-extra-packages =
+    let haskellNix = import sources."haskell.nix";
+    in with (import sources.nixpkgs {
+    inherit system crossSystem;
+    config = haskellNix.config // config;
+    overlays = haskellNix.overlays ++ overlays.haskell-nix-extra;
+  }); {
+    inherit
+      regenerateStackPackages
```
angerman: Why is this not `stackProjectRegenerate` to be in line with `cabalProjectRegenerate`?

default.nix
```diff
@@ -131,6 +143,7 @@ let
       shell
       tests
       rust-packages
+      haskell-nix-extra-packages
```
angerman: Why is this exported via haskell-nix-extra-packages.stackProjectRegenerate when `cabalProjectRegenerate` is exported in the flat name space?

default.nix
jbgi: There was indeed an old `cabal-project-regenerate.nix` file but I just deleted it (not used). 
The new `cabalProjectRegenerate` is completely different than `regenerateStackPackages`: it only update `--sha256` entries in `cabal.project`.

default.nix
jbgi: `cabalProjectRegenerate` does not make use of `haskell.nix`/ `nix-tools`, hence it is ok to use directly (not through overlay).
Conversely, since `stackProjectRegenerate` make use of `haskell.nix`/ `nix-tools` it should be used in projects through the `haskell-nix-extra` overlay in order for the script to use the same version of `haskell.nix`/ `nix-tools` that the project is using. 

`haskell-nix-extra-packages` is mostly there for the script to be built on hydra.

default.nix
disassembler: please rename to stackProjectRegenerate

PR #352, opened 2020-03-11
Fix wrong path to stack-hpc-coverall in overlay.

jbgi:
Also deduplicate definitions.


nc6:
Can confirm this fixes the situation for `cardano-ledger-specs`

PR #362, opened 2020-03-31
Update nix-tools-regenerate.nix

angerman:
I really don't know what's going in with Nixpkgs here, however it appears that `en_US.UTF-8` is gone. `C.utf8` should however always be there, even if LOCALE_ARCHIVE is missing.


kk-hainq:
May I ask the status of this change, or if you plan to merge it? It seems to resolve locale issues when running `stack-to-nix` on a few machines for us.

rvl (approved):

PR #366, opened 2020-04-02
checkStackProject: ignore whitespace in diff

rvl:
As mentioned by @jonathanknowles - it's annoying when this buildkite CI step fails because of a difference in whitespace at the end of the nix file.

### DIFF
--- a/ci/check-stack-project.nix
+++ b/ci/check-stack-project.nix
@@ -24,7 +24,7 @@ writeScript "check-stack-project.sh" ''
   git add -A nix
 
   # Check if there are changes staged for commit.
-  if git diff-index --cached --quiet HEAD --; then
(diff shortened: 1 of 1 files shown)


ci/check-stack-project.nix
```diff
@@ -24,7 +24,7 @@ writeScript "check-stack-project.sh" ''
   git add -A nix
 
   # Check if there are changes staged for commit.
-  if git diff-index --cached --quiet HEAD --; then
+  if git diff-index --ignore-all-space --cached --quiet HEAD --; then
```
jonathanknowles, line 27: Perhaps the `ignore-space-at-eol` option would be more specific here?

rvl: It might be, but since whitespace doesn't matter to nix, I go with the simpler `git diff-index -w` option.

PR #379, opened 2020-05-12
remove skeleton project

disassembler:
We have 20 PRs open and almost all of them have skeleton eval failures that burden CI with constantly trying to re-evaluate and fail over and over again. This will at least help take that burden off. All skeleton files have been moved to cardano-skeleton.


angerman:
I believe this was done.

PR #385, opened 2020-05-27
Fix syntax error in check-stack-project

rvl:
Tested with:

    nix-build -A checkStackProject -o ./check-stack-project.sh && ./check-stack-project.sh

What happened to the CI on this repo?!?


jbgi:
skeleton is now at https://github.com/input-output-hk/cardano-skeleton
we still need to change iohk-nix ci to use it.

PR #388, opened 2020-06-03
use nixpkgs 20.03 by default

jbgi:
also removed skeleton.
jormungandr still built with 19.09 though.


default.nix
```diff
@@ -116,7 +116,7 @@ let
       (commonLib.pkgsDefault.callPackage ./overlays/rust/mozilla.nix {})
       (import ./overlays/rust)
     ];
-    pkgs = import sources.nixpkgs {
+    pkgs = import sources."nixpkgs-19.09" {
```
manveru, line 119: When can we get rid of that?

manveru: I guess when we bump Jormungandr :)

jbgi: I think `nixpkgs-mozilla` rust overlay is not compatible with 20.03. Did not dig deep into the error though.

PR #390, opened 2020-06-10
Respect recurseForDerivations = false

hamishmack:
`recRecurseIntoAttrs` sets `recurseForDerivations = true` everything in
the tree until it finds a derivation, but with the `project` and
`package` attributes added recently to haskell.nix the tree can be
infinite.  For instance you can wind up with something like

native.haskellPackages.contra-tracer.project.hsPkgs.Cabal.project.
hsPkgs.Cabal.components.tests.check-tests.x86_64-linux

To fix this we have added explicit `recurseForDerivations = false`
to the attributes that point back to the `project` and `package`.

However currently `recRecurseIntoAttrs` does not respect that and
turns everything back to `recurseIntoAttrs = true`.

This change updates `recRecurseIntoAttrs` so it will stop whenever it
sees `recurseIntoAttrs = false`.

### DIFF
--- a/overlays/haskell-nix-extra/default.nix
+++ b/overlays/haskell-nix-extra/default.nix
@@ -6,7 +6,7 @@ pkgs: super: with pkgs; with lib; {
          // { recurseForDerivations = true; };
 
       recRecurseIntoAttrs = x:
-        if (isAttrs x && !isDerivation x)
+        if (isAttrs x && !isDerivation x && x.recurseForDerivations or true)
         then recurseIntoAttrs (mapAttrs (n: v: if n == "buildPackages" then v else recRecurseIntoAttrs v) x)
         else x;
 


michaelpj:
Can we avoid the tree being infinite? That does make it hard to process it in lots of ways, e.g. filtering stuff out.

jbgi:
bors merge

hamishmack:
> Can we avoid the tree being infinite? That does make it hard to process it in lots of ways, e.g. filtering stuff out.

I can't think of a good way to do that.  We could change [addProjectAndPackageAttrs](https://github.com/input-output-hk/haskell.nix/blob/da9c5b93a983b03ed2c32c081915dda95cc6917c/overlays/haskell.nix#L478) to only add the input versions of `project` and `package`.  That would mean that `native.haskellPackages.contra-tracer.project.hsPkgs.Cabal` would not have a `project` attribute and the tree would be limited in size.  I'm not keen on this idea though because:

* It would mean only `native.haskellPackages.Cabal` could be used with functions that wanted access to the `.project` (since `native.haskellPackages.contra-tracer.project.hsPkgs.Cabal` would not have the attribute).

* If we are looking into `.project` when filtering something like `native.haskellPackages` then it is probably a bug already and while it may no longer result in an overflow of the stack it is still likely to result in many unwanted duplicates.

michaelpj:
I missed the point where we added `.project` to derivations. I would probably have suggested we just not do that, partly for this reason. It's not so bad to get the project yourself. But I can see why you want it. If you're having references back "up" the expression tree then I think you're just stuck with having an infinite structure.

PR #398, opened 2020-06-22
New configs for shelley_testnet and shelley_qa

disassembler:


cardano-lib/shelley-testnet-config.nix
```diff
@@ -0,0 +1,34 @@
+##########################################################
+###############           Alpha1           ###############
```
johnalotoski, line 2: Is the alpha1 label still wanted?

PR #411, opened 2020-07-20
Support testnet and mainnet for hfc

disassembler:


disassembler:
> This may not be correct?
> 
> `ShelleyGenesisFile = networkConfig.ByronGenesisFile or null;`
> 
> Also, the testnet epoch length here seems to be 5 days rather than 6 hours (and some of the other parameters may also have changed in line with this - I'm not sure whether that's intentional or accidental).

cardano-testnet has always been 5 days. The line mentioned above was a WIP refactoring to make db-sync work with the same config file. It's correct now (and working).

cardano-lib/default.nix
```diff
@@ -23,6 +23,15 @@ let
...
   defaultExplorerLogConfig = import ./explorer-log-config.nix;
+  mkExplorerConfig = name: networkConfig: lib.filterAttrs (k: v: v != null) {
+    NetworkName = name;
+    Protocol = networkConfig.Protocol;
+    RequiresNetworkMagic = networkConfig.RequiresNetworkMagic;
+    ByronGenesisFile = networkConfig.ByronGenesisFile or null;
+    ByronGenesisHash = networkConfig.ByronGenesisHash or null;
+    ShelleyGenesisFile = networkConfig.ShelleyGenesisFile or null;
+    ShelleyGenesisHash = networkConfig.ShelleyGenesisHash or null;
+  };
```
cleverca22, line 34: for the keys where the name matches up, you could `inherit (networkConfig) Protocol;`

cardano-lib/mainnet-shelley-genesis.json
```diff
@@ -0,0 +1,73 @@
...
+    "60baee25cbc90047e83fd01e1e57dc0b06d3d0cb150d0ab40bbfead1": {
+      "delegate": "7f72a1826ae3b279782ab2bc582d0d2958de65bd86b2c4f82d8ba956",
+      "vrf": "c0546d9aa5740afd569d3c2d9c412595cd60822bb6d9a4e8ce6c43d12bd0f674"
+    }
+  },
+  "updateQuorum": 5,
+  "networkId": "Mainnet",
+  "initialFunds": {},
+  "maxLovelaceSupply": 45000000000000000,
+  "protocolMagicId": 764824073,
```
mrBliss, line 61: This field is no longer used.

cardano-lib/mainnet-shelley-genesis.json
```diff
@@ -0,0 +1,73 @@
...
+    },
+    "60baee25cbc90047e83fd01e1e57dc0b06d3d0cb150d0ab40bbfead1": {
+      "delegate": "7f72a1826ae3b279782ab2bc582d0d2958de65bd86b2c4f82d8ba956",
+      "vrf": "c0546d9aa5740afd569d3c2d9c412595cd60822bb6d9a4e8ce6c43d12bd0f674"
+    }
+  },
+  "updateQuorum": 5,
+  "networkId": "Mainnet",
+  "initialFunds": {},
+  "maxLovelaceSupply": 45000000000000000,
```
mrBliss, line 60: Is the same as for Byron?

cardano-lib/mainnet-config.nix
```diff
@@ -6,7 +6,10 @@
 {
   ##### Locations #####
 
-  GenesisFile = ./mainnet-byron-genesis.json;
+  ByronGenesisFile = ./mainnet-byron-genesis.json;
+  ByronGenesisHash = "5f20df933584822601f9e3f8c024eb5eb252fe8cefb24d1317dc3d432e940ebb";
+  ShelleyGenesisFile = ./mainnet-shelley-genesis.json;
+  ShelleyGenesisHash = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
```
mrBliss, line 12: Is this field read? We should use it to check whether the right (unmodified) genesis file is being used.

cardano-lib/mainnet-shelley-genesis.json
```diff
@@ -0,0 +1,73 @@
...
+  "protocolMagicId": 764824073,
+  "networkMagic": 764824073,
+  "epochLength": 432000,
```
mrBliss, line 63: Is this right, @edsko?

cardano-lib/mainnet-shelley-genesis.json
JaredCorduan: good question! yes and no (but this value is correct). Byron doesn't have such explicit accounting numbers for every lovelace in the system. there is currently about 30 billion (billion with nine zeros) ada in UTxO, and as far as I know there has always been a clear and open promise that 45 billion ada is the max. Note that there is one million (six zeros) lovelace in one ada. I just counted the zeros above, there are fifteen. but please double check me :)

cardano-lib/mainnet-shelley-genesis.json
```diff
@@ -0,0 +1,73 @@
...
+    },
+    "maxTxSize": 16384,
+    "maxBlockBodySize": 65536,
+    "maxBlockHeaderSize": 1400,
+    "minFeeA": 44,
+    "minFeeB": 155381,
+    "minUTxOValue": 1000000,
+    "poolDeposit": 500000000,
+    "minPoolCost": 228000000,
+    "keyDeposit": 400000,
```
JaredCorduan, line 21: I don't think this field, `keyDeposit`, should be around still.

cardano-lib/mainnet-shelley-genesis.json
```diff
@@ -0,0 +1,73 @@
...
+  "maxLovelaceSupply": 45000000000000000,
+  "protocolMagicId": 764824073,
+  "networkMagic": 764824073,
+  "epochLength": 432000,
+  "staking": {
+    "pools": {},
+    "stake": {}
+  },
+  "systemStart": "2017-09-23T21:44:51Z",
+  "slotsPerKESPeriod": 129600,
```
JaredCorduan, line 69: did we want 36 hours per KES period?

cardano-lib/mainnet-shelley-genesis.json
JaredCorduan: fwiw, this number is what I expected. it is five days, and `10*(k/f)`, where `k` is called `securityParam` here and `f` is called `activeSlotsCoeff`. which makes our stability window one day.

edsko: No idea :) Question for @JaredCorduan .

cardano-lib/mainnet-shelley-genesis.json
edsko: I think that's what @dcoutts said, yes.

PR #437, opened 2020-09-18
Update Haskell.nix and Nixpkgs match latest

hamishmack:
We have updated the nixpkgs 20.03 commit that we are using for
haskell.nix CI.  This change would update both haskell.nix and
nixpkgs in this repo to match (should help avoid cache misses on
hydra).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,10 @@
 Please read these notes when updating your project's `iohk-nix`
 version. There may have been changes which could break your build.
 
+## 2020-10-06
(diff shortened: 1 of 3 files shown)


rvl (approved):
I am not sure what this is going to break.
The only thing here remaining that uses haskell.nix directly is stackNixRegenerate (requires nix-tools).

The collectChecks function can be removed from iohk-nix now.

rvl:
I tried using this branch on cardano-wallet, but it wouldn't evaluate - [cardano-wallet-pr-2124-eval.txt](https://github.com/input-output-hk/iohk-nix/files/5331228/cardano-wallet-pr-2124-eval.txt).

PR #441, opened 2020-10-09
[SRE-141] Enable hlint of sub-directories

craigem:
Hydra is currently unable to `hlint` projects in sub-directories because it consumes `selectProjectPackages` which only provides a package name which `hlint` then assumes is in the root.

This is an interim workaround.

### DIFF
--- a/tests/hlint.nix
+++ b/tests/hlint.nix
@@ -1,4 +1,4 @@
-{ runCommand, hlint, src, lib, projects }:
+{ runCommand, hlint, src, lib, }:
 
 let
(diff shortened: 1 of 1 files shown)


jbgi:
I see. Can you remove the `projects` arg altogether, since it's not used anymore? Users can use the `src` argument to select projects if needed.

jbgi (changes requested):
I see. Can you remove the projects arg altogether, since it's not used anymore? Users can use the src argument to select projects if needed.

sevanspowell (approved):

PR #444, opened 2020-10-13
Add fix-stylish-haskell script

jbgi:
slightly adapted from plutus repo.


tests/fix-stylish-haskell.nix
```diff
@@ -0,0 +1,20 @@
+{ writeScriptBin, git, fd, stylish-haskell }:
+
+writeScriptBin "fix-stylish-haskell" ''
+    ${git}/bin/git diff > pre-stylish.diff
+    ${fd}/bin/fd \
+      --extension hs \
+      --exclude '*/dist/*' \
+      --exclude '*/docs/*' \
```
michaelpj, line 8: This is project-specific. Maybe you want to take an ignore list as an argument?

michaelpj (approved):
Enjoy ;)

tests/fix-stylish-haskell.nix
rvl, line 8:
```suggestion
    ${fd}/bin/fd --extension hs "$@" \
```
This would be a cheap and cheerful way to set the ignore list.
But the existing excludes would be fine for cardano-wallet at least.

rvl (approved):

PR #448, opened 2020-11-02
Drop support for LiveView

erikd:
LiveView is no longer supported by the node itself.

### DIFF
--- a/cardano-lib/generic-log-config.nix
+++ b/cardano-lib/generic-log-config.nix
@@ -1,10 +1,5 @@
 {
 
-  # The node can run in either the SimpleView or LiveView. The SimpleView just
-  # uses standard output, optionally with log output. The LiveView is a text
-  # console with a live view of various node metrics.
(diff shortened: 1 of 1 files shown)


dcoutts (approved):
LGTM

sevanspowell (approved):

cardano-lib/generic-log-config.nix
```diff
@@ -176,35 +171,11 @@
...
-         "EKGViewBK"
-         {
-            kind = "UserDefinedBK";
-            name = "LiveViewBackend";
-         }
-      ];
+      "cardano.node.ChainDB.metrics" = [];
+      "cardano.node.metrics" = [];
+      "cardano.node-metrics" = [];
       "cardano.node.Forge.metrics" = [ "EKGViewBK" ];
```
erikd, line 177: @sevanspowell  that "EKGViewBK" should also be removed.

PR #449, opened 2020-11-02
Don't set the endpoint path in smashUrl wrt ADP-502

hasufell:
depends on https://github.com/input-output-hk/cardano-wallet/pull/2285


hasufell:
@rvl

rvl:
I'm not sure how these settings are used.
All of the URLs return 404.
Are they imported by the daedalus installer @disassembler ?

hasufell:
@rvl try `curl https://smash.cardano-mainnet.iohk.io/api/v1/status`

PR #452, opened 2020-11-23
cardano-node: add basicInfo to default log config

disassembler:


jbgi (approved):

deepfire:
This is not actually needed. `cardano_node_metrics_nodeStartTime_int` is the new metric implemented in https://github.com/input-output-hk/cardano-node/pull/2118 that works with the default config.

PR #477, opened 2021-05-26
support alonzo for all environments

disassembler:


cardano-lib/mainnet-config.nix
```diff
@@ -6,10 +6,12 @@
 {
   ##### Locations #####
 
-  ByronGenesisFile = ./mainnet-byron-genesis.json;
+  ByronGenesisFile = ./mainnet + "/byron-genesis.json";
```
manveru, line 9: why not just `./mainnet/byron-genesis.json`?

cleverca22 (approved):

cardano-lib/mainnet-config.nix
disassembler: because I want all the files in the same store path. Main reason is alonzo genesis *may* need to contain a reference to another json file relative to where it is.

johnalotoski (approved):

PR #489, opened 2021-07-27
nix-shell/cabalWrapped: use separate cabal.project

jbgi:
to avoid modifying original cabal.project.

### DIFF
--- a/overlays/utils/default.nix
+++ b/overlays/utils/default.nix
@@ -16,16 +16,13 @@ in {
     GC_DONT_GC=1 ${nixFlakes}/bin/nix --experimental-features "nix-command flakes" "$@"
   '';
   cabalWrapped = writeShellScriptBin "cabal" ''
-    >&2 echo 'Temporary modify `cabal.project` to use nix builds of `source-repository-package`s (haskell/cabal#5444 and haskell/cabal#6249).'
-    PROJECT="$(${git}/bin/git rev-parse --show-toplevel)/cabal.project"
-    BACKUP="$(mktemp)"
-    cp -a "$PROJECT" "$BACKUP"
-    sed -ni '1,/--- 8< ---/ p' $PROJECT
-    function atexit() {
-        mv "$BACKUP" "$PROJECT"
-    }
-    trap atexit EXIT
-    ${final.cabal or final.cabal-install}/bin/cabal "$@"
+    TOPLEVEL="$(${git}/bin/git rev-parse --show-toplevel)"
+    ORIG_PROJECT="$TOPLEVEL/cabal.project"
+    # Should be in .gitignore:
+    NIXSHELL_PROJECT="$TOPLEVEL/.nix-shell-cabal.project"
+    
+    sed -n '1,/--- 8< ---/ p' $PROJECT > "$NIXSHELL_PROJECT"
+    exec ${final.cabal or final.cabal-install}/bin/cabal --project-file="$NIXSHELL_PROJECT" "$@"
   '';
 
   hlintCheck = ../../tests/hlint.nix;
--- a/overlays/utils/default.nix
+++ b/overlays/utils/default.nix
@@ -16,13 +16,51 @@ in {
     GC_DONT_GC=1 ${nixFlakes}/bin/nix --experimental-features "nix-command flakes" "$@"
   '';
   cabalWrapped = writeShellScriptBin "cabal" ''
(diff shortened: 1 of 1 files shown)


overlays/utils/default.nix
```diff
@@ -16,16 +16,13 @@ in {
...
-        mv "$BACKUP" "$PROJECT"
-    }
-    trap atexit EXIT
-    ${final.cabal or final.cabal-install}/bin/cabal "$@"
+    TOPLEVEL="$(${git}/bin/git rev-parse --show-toplevel)"
+    ORIG_PROJECT="$TOPLEVEL/cabal.project"
+    # Should be in .gitignore:
+    NIXSHELL_PROJECT="$TOPLEVEL/.nix-shell-cabal.project"
+    
+    sed -n '1,/--- 8< ---/ p' $PROJECT > "$NIXSHELL_PROJECT"
```
rvl, line 24: Should it be ORIG_PROJECT instead of PROJECT?

And can we please use a more clever filter? Like this:
```suggestion
    awk '
  # Matches all section starts
  /^[^ ]/ {
    # Remember the section name (they can span multiple lines)
    section = $0
  }
  # Matches every line
  // {
    # Only print the line if it is not in the section we want to omit
    if (section != "source-repository-package")
      print $0
  }
' "$ORIG_PROJECT" > "$NIXSHELL_PROJECT"
```

overlays/utils/default.nix
```diff
@@ -16,16 +16,13 @@ in {
...
+    exec ${final.cabal or final.cabal-install}/bin/cabal --project-file="$NIXSHELL_PROJECT" "$@"
```
rvl, line 25: Can we log the real cabal command to stderr? Otherwise users may get confused by the wrapper.

```suggestion
    >&2 echo cabal blah blah
    exec ${final.cabal or final.cabal-install}/bin/cabal --project-file="$NIXSHELL_PROJECT" "$@"
```

overlays/utils/default.nix
```diff
@@ -16,16 +16,13 @@ in {
...
-    PROJECT="$(${git}/bin/git rev-parse --show-toplevel)/cabal.project"
-    BACKUP="$(mktemp)"
-    cp -a "$PROJECT" "$BACKUP"
-    sed -ni '1,/--- 8< ---/ p' $PROJECT
-    function atexit() {
-        mv "$BACKUP" "$PROJECT"
-    }
-    trap atexit EXIT
-    ${final.cabal or final.cabal-install}/bin/cabal "$@"
+    TOPLEVEL="$(${git}/bin/git rev-parse --show-toplevel)"
```
rvl, line 19: It's possible to find the top-level in bash without `git`, and without the assumption that `cabal.project` sits at the repo top-level.

Also it would be more failsafe to bypass wrapping if the toplevel couldn't be found.

rvl (approved):
Thanks, @jbgi, I like these changes. I have a suggestion for a better wrapper though:

```nix
writeShellScriptBin "cabal" ''
  set -euo pipefail

  find_up() {
    while [[ $PWD != / ]] ; do
      if [[ -e "$1" ]]; then
        echo "$PWD"
        return
      fi
      cd ..
    done
  }

  toplevel=$(find_up "cabal.project")

  if [[ -n "$toplevel" ]]; then
    nix_shell_cabal_project=$toplevel/.nix-shell-cabal.project
    extra_cabal_opts=("--project-file=$nix_shell_cabal_project")
    awk '
      # Matches all section starts
      /^[^ ]/ {
        # Remember the section name (they can span multiple lines)
        section = $0
      }
      # Matches every line
      // {
        # Only print the line if it is not in the section we want to omit
        if (section != "source-repository-package")
          print $0
      }
    ' "$toplevel/cabal.project" > "$nix_shell_cabal_project"
  else
    extra_cabal_opts=()
  fi

  cabal="${final.cabal or final.cabal-install}/bin/cabal"

  >&2 echo "$cabal ''${extra_cabal_opts[@]} $@"
  exec "$cabal" "''${extra_cabal_opts[@]}" "$@"
''
```

jbgi:
@rvl your suggested wrapper is indeed much better. Feel free to open an new PR or modify this one.

rvl (approved):

rvl:
OK done, thanks @jbgi

rvl:
I just realised that for my `shell.nix` it would be better to have a `makeCabalWrapper` function that has the cabal package as its function argument.

PR #501, opened 2021-11-04
Add sha256map-regenerate.

jbgi:


ci/sha256map-regenerate/sha256map-regenerate.py
```diff
@@ -0,0 +1,51 @@
...
+  dict = match.groupdict()
+  prefetchJSON = subprocess.run(
+    ["nix-prefetch-git", "--fetch-submodules", "--quiet", dict['loc'], dict['tag']],
+    capture_output=True, check=True).stdout
+  sha256 = json.loads(prefetchJSON)["sha256"]
+  return """
+  "{loc}"."{tag}" = "{sha256}";
+""".format(**{**dict, **{"sha256": sha256}})
+
+f = open("nix/sha256map.nix",'w')
```
rvl, line 44: It's probably easier and more flexible to use `sys.stdout` and `sys.stdin` for input and output files.
Or even use the [argparse](https://docs.python.org/3/library/argparse.html) module, with `argparse.FileType` (super easy).

rvl (approved):
Sweet. Luckily we don't use a custom stack resolver URL any more.

ci/sha256map-regenerate/default.nix
```diff
@@ -0,0 +1,11 @@
...
+, runtimeShell
+, lib
+, python37
+, nix-prefetch-git
+}:
+writeScript "sha256map-regenerate.sh" ''
+  #!${runtimeShell}
+  export PATH="${lib.makeBinPath [ nix-prefetch-git python37 ]}"
+  ${python37}/bin/python ${./sha256map-regenerate.py}
+''
```
rvl, line 11: Here's one way I like to do PATH wrapping with scripts (shell or python):

```suggestion
{ stdenv
, lib
, python37
, nix-prefetch-git
}:
stdenv.mkDerivation {
  name = "sha256map-regenerate";
  src = ./sha256map-regenerate.py;
  dontUnpack = true;
  dontConfigure = true;
  dontBuild = true;
  installPhase = ''
    bin="$out/bin/${name}"
    sed $src \
      '2i import sys; ${lib.concatMapStringsSep ";" (p: "sys.path.insert(0, \"${p}/bin\")") [ nix-prefetch-git python37 ]}' \
      "$bin"
    chmod 755 "$bin"
    patchShebangs "$bin"
  '';
}
```
(something like it anyway - I haven't tested this example)

rvl (approved):

PR #502, opened 2021-12-14
Bump and clean-up niv pins / ghc.

jbgi:


default.nix
```diff
@@ -113,16 +111,6 @@ let
...
-  rust-packages = rec {
-    overlays = [
-      (commonLib.pkgsDefault.callPackage ./overlays/rust/mozilla.nix {})
-      (import ./overlays/rust)
-    ];
-    pkgs = import sources."nixpkgs-19.09" {
-      inherit overlays config system crossSystem;
-    };
-  };
-
```
disassembler, line 125: We're using rust-packages with some projects utilizing naersk. This needs to stay. Ideally we should manage the mozilla overlay with flakes though.

default.nix
```diff
@@ -150,10 +138,8 @@ let
       sources
       shell
       tests
-      rust-packages
       haskell-nix-extra-packages
-      cardanoLib
-      jormungandrLib;
```
disassembler, line 156: good riddance!

disassembler (changes requested):

PR #503, opened 2022-02-09
[WIP] Update haskell.nix

kozross:
This is necessary for [this issue](https://github.com/input-output-hk/plutus/issues/4233), specifically for the support of Schnorr signatures, as per [this PR](https://github.com/input-output-hk/plutus/pull/4368#issuecomment-1031782137). As all our SECP256k1 support currently goes via `cardano-base` (see [this PR](https://github.com/input-output-hk/cardano-base/pull/252)), we need access to an updated `libsecp256k1`, which has only been added to `nixpkgs-unstable` [recently](https://github.com/NixOS/nixpkgs/pull/158516).

As this currently depends on [this PR](https://github.com/input-output-hk/haskell.nix/pull/1368) landing, I have marked it WIP. Once the `haskell.nix` PR lands, I will update to aim at `haskell.nix` `master` instead of our fork.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,10 @@
 Please read these notes when updating your project's `iohk-nix`
 version. There may have been changes which could break your build.
 
+## 2022-02-10
+
+  * Bump `haskell.nix`.
+
 ## 2021-07-22
   * Renamed `libsodium` to `libsodium-vrf` in the crypto overlay. This
     allows much more sharing from the binary caches.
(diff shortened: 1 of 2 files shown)


michaelpj:
You could also just teach `cardano-base` to pass an explicit nixpkgs pin instead of taking the one from `iohk-nix`.

L-as:
@michaelpj Isn't it best to use a consistent Nixpkgs pin across the projects?

PR #511, opened 2022-05-10
cardano-lib:  introduce the modern tracing log config;  retain the legacy as legacyLogConfig

deepfire:


cardano-lib/generic-log-config.nix
```diff
@@ -7,208 +7,80 @@
...
-    "KatipBK"
-    # EKGViewBK
+  # Use the modern tracing system
+  UseTraceDispatcher = true;
+
+  # Set severities for the tracing namespace
+  TraceOptionSeverity = [
+
+    # Globally, i.e. default:
+    { ns = "";                      severity = "Notice"; };
```
michalrus, line 17: How can this `;` at the end be parsed? :open_mouth:

deepfire: Oh, wow -- thank you for catching this @michalrus !

PR #514, opened 2022-08-09
Use a lib sub-directory to bundle macos dyn libs.

jbgi:
so that they are not mixed with executables,
 thus avoiding colision issues, eg., with devshell.


jbgi:
not sure this is the best way though. probably better to keep `rewrite-libs` as is but use it only for release bundles.

PR #522, opened 2023-01-31
Reinstate proofbytes for backwards compatibility

iquerejeta:


angerman (approved):

angerman:
I want to try this against node... but compilation cycles are _insane_, especially as we have to consistently rebuild it all.

angerman:
tested with db-sycn.

PR #523, opened 2023-02-18
Sort sha256map entries.

philderbeast:
Sorts the sha256map entries. Useful if storing in source control.


angerman:
LGTM. Sorry this took so long.

PR #525, opened 2023-02-22
haskell-nix-extra overlay: add generic setGitRev

jbgi:
also removed old deprecated haskellLib.extra


angerman (approved):
LGTM. Comment is a bit light though.

overlays/haskell-nix-extra/default.nix
```diff
@@ -40,4 +24,30 @@ in final: prev: with final; with lib; {
       chmod -R +w $out
       $STRIP $out/bin/*
     '' else p;
+
+  # Stamp one executable with version info.
```
shlevy, line 28: What does it mean to "stamp" an executable?

jbgi: providing the git rev to be printed with `--version`
(added some more doc)

PR #528, opened 2023-03-09
Iquerejeta/include blst

iquerejeta:
Include the BLST library as part of iohk-nix. This is required for the effort of including BLS12-381 operations in plutus: 
* https://github.com/input-output-hk/cardano-base/pull/266

### DIFF
--- a/overlays/crypto/blst.nix
+++ b/overlays/crypto/blst.nix
@@ -12,10 +12,8 @@ stdenv.mkDerivation rec {
 
   installPhase = ''
     ./build.sh
-    mkdir -p $out/bin
-    cp ./libblst.a $out/bin/libblst.a
-    chmod +x $out/bin/libblst.a
-    ls $out/bin/
+    mkdir -p $out/lib
+    cp ./libblst.a $out/lib/libblst.a
   '';
 
   enableParallelBuilding = true;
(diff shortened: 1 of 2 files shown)


michaelpj:
Is this not in `nixpkgs`? Can we upstream it to there?

iquerejeta:
I thought it'd make sense to have it here with the rest of the crypto libraries.

michaelpj:
Well, we have `libsodium` here because it's an IOHK fork of upstream, and so there's no way we could get anyone else to distribute it. But this seems to be a totally legit normal C library with no funny stuff, so why not just package it in `nixpkgs`? That might take a while to trickle through, so we might want to do it here as well in the mean time.

angerman:
I think @michaelpj raises an important point. We *should* upstream this. 

There is however two issues we need to be aware of. 
- upstreaming will take time and even if we get a backport-22.11, it will take at least 2wks, and we would need to bump nixpkgs everywhere.
- once upstreamed, we'd need to periodically keep an eye on it, to ensure no one breaks it in nixpkgs.

So for now, I'm ok with it in iohk-nix.

angerman:
@iquerejeta can you also prepare an upstream PR? @yvan-sraka can probably help you if needed.

iquerejeta:
Great, will prepare a PR in upstream. Will ping here when opened. Can we merge in the meantime?

PR #529, opened 2023-04-04
dynamic blst

angerman:


overlays/crypto/blst.nix
```diff
@@ -10,10 +11,37 @@ stdenv.mkDerivation rec {
...
+      fi
+    done
+  '' + ''
+    mkdir -p $out/lib/pkgconfig
+    cat <<EOF > $out/lib/pkgconfig/libblst.pc
+    prefix=$out
+    exec_prefix=''\\''${prefix}
+    libdir=$out/lib
+
+    Name: libsodium
```
lehins, line 39: Should this be `blst` instead of `sodium`?
```suggestion
    Name: libblst
```

overlays/crypto/blst.nix
```diff
@@ -2,6 +2,7 @@
 
 stdenv.mkDerivation rec {
   name = "blst-0.3.10";
```
andreabedini, line 4: `name` is set by default to `${pname}-${version}" so you can just do `pname = "blst"` here.

overlays/crypto/blst.nix
```diff
@@ -10,10 +11,45 @@ stdenv.mkDerivation rec {
...
+    exec_prefix=''\\''${prefix}
+    libdir=$out/lib
+
+    Name: libblst
+    Version: ${version}
+    Description: ${meta.description}
+
+    Libs: -L''\\''${libdir} -lblst
+    EOF
+  '';
```
andreabedini, line 45: How much of this is done by `installPhase` already? could you get away with adding something to `postInstallPhase` (e.g. the pkgconfig stuff?).

andreabedini (approved):

overlays/crypto/blst.nix
angerman: Nothing really. There is zero install logic in the repo. All they have is this very basic setup.sh script: https://github.com/supranational/blst/blob/master/build.sh

overlays/crypto/blst.nix
```diff
@@ -10,10 +11,45 @@ stdenv.mkDerivation rec {
...
+    cat <<EOF > $out/lib/pkgconfig/libblst.pc
+    prefix=$out
+    exec_prefix=''\\''${prefix}
+    libdir=$out/lib
+
+    Name: libblst
+    Version: ${version}
+    Description: ${meta.description}
+
+    Libs: -L''\\''${libdir} -lblst
```
michaelpj, line 43: should this be `-llibblst`? Just wondering if we have some name confusion between blst and libblst or something.

PR #533, opened 2023-04-25
[GHA] for iohk-nix

angerman:
- [x] build and install `libsodium`
- [x] build and install `secp256k1`
- [x] build and install `libblst`
- [x] pre-build binaries
- [x] pull binaries from GH artifacts


action.yml
```diff
@@ -0,0 +1,90 @@
+name: 'Setup IOG librararies'
+description: 'Installs libsodum, libsecp256k1 and libblst'
+inputs:
```
michaelpj, line 3: dead

action.yml
```diff
@@ -0,0 +1,90 @@
+name: 'Setup IOG librararies'
```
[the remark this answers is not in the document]
angerman: They really annoying issue with GHAs is that you have action@tag. And you can't really tag different folders :-/

But yes, maybe moving them into iog/actions/{base,...} might be a good idea.

I had them in iohk-nix, as I wanted to have some consistency between iohk-nix, and the GitHub Actions.

action.yml
angerman: Right. We probably won't need this anymore. I wasn't sure in the beginning.

action.yml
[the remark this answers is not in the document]
michaelpj: It did seem a bit weird having the actions here, but given that you've got the machinery for pre-buiding them here it maybe makes sense? IDK. You're going to have some kind of cross-repo dependency to point to where the pre-built artifacts will be anyway, right?

.github/workflows/release.yml
```diff
@@ -0,0 +1,44 @@
+name: Release
```
michaelpj, line 1: It also seems kind of weird for this to be here, but IDK where else it could go.

flake.nix
```diff
@@ -28,6 +30,234 @@
     };
 
     pkgs = import nixpkgs { system = "x86_64-linux"; overlays = builtins.attrValues overlays; };
+    darwin-pkgs = import nixpkgs { system = "x86_64-darwin"; overlays = builtins.attrValues overlays; };
+
+
+    # we can use this, to get a coherent picture of the sources for
```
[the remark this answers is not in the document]
angerman: Because others want to know :-)

flake.nix
```diff
@@ -28,6 +30,234 @@
...
+    lib-srcs = {
+      secp256k1 = pkgs.secp256k1.src.url;
+      sodium    = pkgs.libsodium-vrf.src.url;
+      blst      = pkgs.libblst.src.url;
+    };
+
+    dist = let
+      # For packaging, we can'd deal with split outputs.
+      mkSingleOutput = drv: drv.overrideDerivation (drv': { outputs = [ "out" ]; });
+      mkDebianPkg = prefix: drv: let
```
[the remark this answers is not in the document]
angerman: The best I know is the nixpkgs infrastructure which _builds_ Debian packages with debutils. I don't know about any lightweight packaging solution that _just_ creates the packages.

.github/workflows/release.yml
angerman: Yes, I've been debating this _a lot_ as well. iohk-nix is the ground truth for our libraries. As this is basically just taking what's _in_ the repo, and publishing it on GH. I wouldn't know how else to do it, other than starting to work this in hydra pushing stuff to GH, which is going to be messy. And then hydra specific :-/

flake.nix
```diff
@@ -28,6 +30,234 @@
...
+
+          # make it downloadable from hydra.
+          mkdir -p $out/nix-support
+          for f in $out/*.deb; do
+            echo "file binary-dist \"''${f}\"" \
+                >> $out/nix-support/hydra-build-products
+          done
+        '';
+      };
+      mkDarwinPkg = prefix: drv: let
```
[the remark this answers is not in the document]
angerman: Ultimately we want to extract this into `pkg.nix`.

.github/workflows/github-action.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  integration:
+    strategy:
+        fail-fast: false
+        matrix:
+          os: [ubuntu-latest, windows-latest, macos-latest]
+    name: GitHub Action integration test
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: input-output-hk/actions/base@lastet
```
andreabedini, line 12:
```suggestion
    - uses: input-output-hk/actions/base@latest
```

angerman:
What a tool I am.

andreabedini (approved):
:zap:

angerman:
Neither I nor @andreabedini could find any comparable pre-existing packaging logic.

flake.nix
```diff
@@ -28,6 +30,239 @@
...
+    #
+    lib-srcs = {
+      secp256k1 = pkgs.secp256k1.src.url;
+      sodium    = pkgs.libsodium-vrf.src.url;
+      blst      = pkgs.libblst.src.url;
+    };
+
+    dist = let
+      # For packaging, we can'd deal with split outputs.
+      mkSingleOutput = drv: drv.overrideDerivation (drv': { outputs = [ "out" ]; });
```
hamishmack, line 51: How does this work?  If a derivation's `buildPhase` or something expects there to be multiple outputs like `$out`, `$doc` and `$dev` is it not likely to still attempt to use those undefined env vars with unpredictable results?

angerman: This just overrides the output for the given drv. And yes it only works with the auto-split stuff if someone explicitly creates the $xxx, this will fail.

PR #534, opened 2023-05-04
Drop old envs

angerman:
This should remove all the old environments from iohk-nix, and use the one from cardano-world.

We still have a risk of iohk-nix, not being up-to-date enough with cardano-world.
A GHA might be a solution for that. I'll look into that In a separate PR.

We may also want to just outright deprecate all this stuff, and ask people to use cardano-world,
now that we pull the slew of dependencies from there anyway :(


angerman:
This is currently referenced in [cardano-db-sync#1403](https://github.com/input-output-hk/cardano-db-sync/pull/1403/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R9), and cardano-db-sync will need to be updated _after_ this PR was merged.

cardano-lib/default.nix
```diff
@@ -21,21 +21,78 @@ let
...
     NetworkName = name;
     inherit (nodeConfig) RequiresNetworkMagic;
     NodeConfigFile = "${__toFile "config-${toString name}.json" (__toJSON nodeConfig)}";
   };
 
+  mkDbSyncConfig = name: nodeConfig: (lib.filterAttrs (k: v: v != null) {
+    NetworkName = name;
+    inherit (nodeConfig) RequiresNetworkMagic;
+    NodeConfigFile = "${__toFile "config-${toString name}.json" (__toJSON nodeConfig)}";
+  })
```
angerman, line 78: am I missing something or is
```nix
mkDbSyncConfig = name: nodeConfig: (mkExplorerConfig name nodeConfig) // defaultDbSyncLogConfig
```
?

angerman (commented):
LGTM. I guess the formatting is due to `nixfmt`?

cardano-lib/default.nix
johnalotoski: You're right.  The `defaultDbSyncLogConfig` alias is also unnecessary.

angerman:
LGT Everyone.

PR #543, opened 2023-05-30
move haskell-nix stuff out of crypto overlay

jbgi:
so that `crypto` can be imported before haskell.nix overlay (though not currently necessary, but could be if we modify nixpkgs in  there in the future), and the new `haskell-nix-crypto` overlay being imported after.

### DIFF
--- a/overlays/crypto/default.nix
+++ b/overlays/crypto/default.nix
@@ -1,4 +1,4 @@
-inputs: final: prev: rec {
+inputs: final: prev: {
   # We pin our own crypto libraries here, so that we have control over
   # the specific revisions we use (assuming a recent enough iohk-nix).
   #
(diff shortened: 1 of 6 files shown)


angerman:
@jbgi thanks for this!

overlays/crypto/default.nix
```diff
@@ -10,15 +10,18 @@ inputs: final: prev: rec {
   libblst = final.callPackage ./libblst.nix { inherit inputs; };
   libsecp256k1 = final.callPackage ./libsecp256k1.nix { inherit inputs; };
 
-  # override the nixpkgs ones which do not have the `lib` prefix.
-  secp256k1 = libsecp256k1;
+  # XXX: deprecated: override the nixpkgs ones which do not have the `lib` prefix.
+  secp256k1 = builtins.trace "WARNING: overriding nixpkgs' secp256k1 with iohk-nix libsecp256k1. This override will go away in July 2023, please use libsecp256k1 directly."
+    libsecp256k1;
```
angerman, line 15: This is great! Thank you!

overlays/crypto/libsecp256k1.nix
```diff
@@ -11,6 +11,8 @@ stdenv.mkDerivation rec {
   configureFlags = [
     "--enable-benchmark=no"
     "--enable-module-recovery"
+  ] ++ lib.optional stdenv.hostPlatform.isMusl [
+    "--enable-static"
   ];
```
angerman, line 17: I don't think this is good. We have many more customisations in haskell.nix, that I'm not sure should be in iohk-nix? For `musl` I can see this being on ok.

angerman (commented):
After reviewing this, I think the proper solution here is to stop tangling the overlay (adding packages) to nixpkgs. And the haskell.nix customisation in _one_ overlay.

I think we want _one_ crypto overlay and one haskell.nix overlay. Such that they can be loaded as

```
iohk-nix/crypo # nixpkgs augmentation

haskell.nix
iohk-nix/haskell.nix # haskell.nix augmentation
```

jbgi:
> I think we want _one_ crypto overlay and one haskell.nix overlay. 

I kind-of wanted to do that too (using the existing haskell-nix-extra overlay for the haskell-nix stuff), but then it would have induced a new dependency on the crypto overlay being loaded.

So probably a new `crypto-haskell-nix` (or `haskell-nix-crypto`?) would be the way to go.

overlays/crypto/libsecp256k1.nix
jbgi: well, if we don't want to override nixpkgs `secp256k1` , then we need to implement here the customizations of `secp256k1` done by haskell.nix musl overlay.

angerman: @iquerejeta mentioned that we want to pin our crypto libs to insulate us from upstream changes and have better control over the exact versions.

jbgi: I meant `haskell.nix` is adding `"--enable-static"` to `secp256k1`, but if we use a different name (`libsecp256k1`) then we need to manage `"--enable-static"` in iohk-nix, right?

angerman:
@jbgi how much code can we save if we just blatantly override the nix supplied `secp256k1` instead of having to deal with our own `libsecp256k1`? Yes it's a bit ugly from the naming scheme with the other libraries, but are the hoops we jump through here to deal with `secp256k1` vs `libsecp256k1` worth it?

angerman:
After our discussion yesterday, I think we should make an exception here for secp256k1, and just follow the upstream nix naming. This will make things much easier on us. We should do the same for libsodium, and eventually libblst (however that ends up in nixpkgs).

angerman:
@jbgi let me know what you think. If you agree, this is good to merge I think.

angerman (approved):
LGTM

PR #559, opened 2023-10-04
Add enableShared to libblst.nix

hamishmack:
This also disables shared by default for windows.

When this `blst.dll` is present, the plutus-core build hangs after it is loaded into the TH interpreter.


hamishmack:
Building `plutus-core` also needs https://github.com/input-output-hk/haskell.nix/pull/2080 (since with this PR a `bin` directory is no longer created).

PR #564, opened 2023-11-29
libblst respect enableShared = false also on darwin

angerman:
Currently we expect the dylib to be there irrespective of enableShared or not. However if enableShared = false, this file does not exist, and it will fail to build.


andreabedini (approved):
:+1:

PR #577, opened 2024-04-15
conway: Add an `EnableFutureGenesis` flag for db sync

sgillespie:
This is the first step for: https://github.com/IntersectMBO/cardano-db-sync/issues/1674. Note that cardano-db-sync ignores unknown configuration values, so coordination/timing is not important.

### DIFF
--- a/cardano-lib/default.nix
+++ b/cardano-lib/default.nix
@@ -298,6 +298,9 @@ let
       networkConfig = import ./private-config.nix // minNodeVersion;
       networkConfigBp = import ./private-config-bp.nix // minNodeVersion;
       usePeersFromLedgerAfterSlot = 1886369;
+      extraDbSyncConfig = {
+        enableConway = true;
+      };
     };
   };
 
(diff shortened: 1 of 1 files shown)


cardano-lib/default.nix
```diff
@@ -252,6 +257,9 @@ let
       edgePort = 3001;
       networkConfig = import ./sanchonet-config.nix;
       usePeersFromLedgerAfterSlot = 21599922;
+      extraDbSyncConfig = {
+        enableConway = true;
```
johnalotoski, line 261: private chain, below, would also have enableConway = true.
I also do like the EnableFutureGenesis or similar generalization suggested in the PR header mentioned issue.

kderme: Agreed with `EnableFutureGenesis`. Did the node config have or still has a similar option? It would be good to have consistency in names.

kderme (approved):
LGTM

johnalotoski (approved):

PR #585, opened 2024-07-11
Update sanchonet conway-genesis for respin

johnalotoski:


Crypto2099 (approved):
LGTM

Ryun1 (approved):

kevinhammond:
Note that the DRep voting thresholds both need to be 65%...

“committeeNoConfidence”: 0.6

kevinhammond (changes requested):
The DRep voting thresholds both need to be 65%
`“committeeNoConfidence”: 0.6`

disassembler:
addressed @kevinhammond comments. Also made a bunch of changes with Johnny Kelly where it differed from recommendation.

Purposefully set govActionLifetime *above* the guardrail because of the short epochs
Purposefully set minCommitteeSize to 5 (from recommended 7) because only 5 ICC members were able to provide keys for the respin.

PR #594, opened 2025-03-02
New node tracing system support

johnalotoski:
This PR changes the default tracing system to the new tracing system and implements support for this where needed:

For Cardano-lib:
* Per environment attributes of `nodeConfig` and `nodeConfigBp` now are populated with new tracing system configuration
* New per environment attributes of `nodeConfigLegacy` and `nodeConfigBpLegacy` are available until the legacy tracing system is removed from node
* New per environment attribute of `tracerConfig` is available, for use with cardano-tracer
* New exposed top level attributes include `defaultLogConfigLegacy` and `defaultTracerConfig`
* The `mkConfigHtml` function has been updated to include legacy configuration files, `$ENV-config-legacy.json` and `$ENV-config-bp-legacy.json`, and a new tracer config file, `$ENV-tracer-config.json`

For generic log config:
* `cardano-lib/generic-log-config.nix` now holds new tracing system default config instead of legacy config

For generic tracing config:
* `cardano-lib/generic-tracer-config.nix` now holds tracing system default config

For testnet template config:
* `cardano-lib/testnet-template/config.json` now holds new tracing system template config instead of legacy config

For legacy generic log config:
* `cardano-lib/generic-log-config-legacy.nix` now holds legacy tracing system default config

For legacy testnet template config:
* `cardano-lib/testnet-template/config-legacy.json` now holds legacy tracing system template config

Other changes include:
* The environment attribute `usePeersFromLedgerAfterSlot` has been renamed to `useLedgerAfterSlot` for consistency with the topology file key value it sets.  A corresponding change has been made in the nixos cardano-node service.
* The default auto-generated topology file name has been changed from `topology.yaml` to `topology.json` to reflect json contents.
* Nixpkgs was bumped to a recent version and breakage changes were resolved with overlay adjustments for `bomutils` and `nixWrapped` util.


cardano-lib/generic-tracer-config.nix
```diff
@@ -0,0 +1,53 @@
+{
+  # Cardano-tracer will use default values where parameters are set to null.
+  WarnRTViewMissing = null;
```
mgmeier, line 3: That option doesn't exist.

cardano-lib/generic-tracer-config.nix
```diff
@@ -0,0 +1,53 @@
...
+    epPort = 12788;
+  };
+
+  # Whether to enable a prometheus export of metrics.
+  hasPrometheus = {
+    epHost = "127.0.0.1";
+    epPort = 12808;
+  };
+
+  loRequestNum = null;
```
mgmeier, line 19: Maybe add comment to clarify:
`# How many trace objects will be forwarded with each request`

cardano-lib/generic-tracer-config.nix
```diff
@@ -0,0 +1,53 @@
+{
+  # Cardano-tracer will use default values where parameters are set to null.
+  WarnRTViewMissing = null;
+  ekgRequestFreq = null;
+  ekgRequestFull = null;
```
mgmeier, line 5: Maybe add comment to clarify:
`# Request full set of metrics always, vs. deltas only (safer, but more overhead)`

cardano-lib/generic-tracer-config.nix
```diff
@@ -0,0 +1,53 @@
+{
+  # Cardano-tracer will use default values where parameters are set to null.
+  WarnRTViewMissing = null;
+  ekgRequestFreq = null;
```
mgmeier, line 4: Maybe add comment to clarify:
`# Interval to request metrics forwarding (seconds)`

cardano-lib/generic-tracer-config.nix
```diff
@@ -0,0 +1,53 @@
...
+  # Configure default logging
+  logging = [
+    {
+      logFormat = "ForHuman";
+      logMode = "FileMode";
+      logRoot = "/tmp/cardano-tracer";
+    }
+  ];
+
+  metricsHelp = null;
```
mgmeier, line 30: `# JSON file or object containing a key-value map {"metric_name": "help text", ...} for Prometheus "# HELP " annotations`

cardano-lib/generic-tracer-config.nix
```diff
@@ -0,0 +1,53 @@
...
+  # Configure default cardano-tracer operational mode
+  network = {
+    contents = "/tmp/cardano-tracer.socket";
+    tag = "AcceptAt";
+  };
+
+  # The network magic will need to be set per environment
+  # networkMagic = ... ;
+
+  resourceFreq = null;
```
mgmeier, line 42: `# Have the cardano-tracer process trace its own resource usage every n seconds. (development / testing)`

cardano-lib/generic-tracer-config.nix
```diff
@@ -0,0 +1,53 @@
...
+
+  # Configure default rotation
+  rotation = {
+    rpFrequencySecs = 60;
+    rpKeepFilesNum = 14;
+    rpLogLimitBytes = 10000000;
+    rpMaxAgeHours = 24;
+  };
+
+  verbosity = null;
```
mgmeier, line 52: `# Verbosity of cardano-tracer's own messages`

cardano-lib/generic-tracer-config.nix
```diff
@@ -0,0 +1,53 @@
...
+  logging = [
+    {
+      logFormat = "ForHuman";
+      logMode = "FileMode";
+      logRoot = "/tmp/cardano-tracer";
+    }
+  ];
+
+  metricsHelp = null;
+  metricsNoSuffix = null;
```
mgmeier, line 31: `# Drop metrics name suffixes (like "_int") (Prometheus only)`

cardano-lib/generic-tracer-config.nix
johnalotoski: It appears in the code [here](https://github.com/IntersectMBO/cardano-node/blob/master/cardano-tracer/src/Cardano/Tracer/Run.hs#L63-L71), and also appears in the `10.3.1` cardano-tracer config dump upon startup as:
```
cardano-tracer[$PID]: {
  "data":{
    "Config":{
      "WarnRTViewMissing":false,
      ...
    },
  },
  ...
}
```
I also have it in the nixos service option in the node PR.
Happy to remove it, though, if you think it should be gone.

johnalotoski: Ah, I see, it's outside of the Config object.

johnalotoski: Removed

cardano-lib/generic-tracer-config.nix
johnalotoski: Good idea.  I have fairly detailed descriptions for everything in the nixos tracer service in the node PR, so I have ported all those descriptions over to this generic tracer config with a few tweaks here and there.  I think these new verbose descriptions should cover everything you have suggested in the comments here, so I will mark these comments as resolved, but please feel free to re-flag anything you think is still an issue.

cardano-lib/generic-log-config.nix
```diff
@@ -6,213 +6,135 @@
...
-    # This overrides the defaultBackends listed above. And note that it is
-    # and override and not an extension so anything matched here will not
-    # go to the default backend, only to the explicitly listed backends.
-    mapBackends = {
-      "cardano.node.metrics" = ["EKGViewBK"];
-      "cardano.node.resources" = ["EKGViewBK"];
+  # Tracing options for node
+  TraceOptions = {
+    # The default tracer configuration
+    "" = {
```
angerman, line 46: This empty string feels really weird.

flake.nix
```diff
@@ -2,7 +2,7 @@
   description = "IOHK nix lib, packages and overlays";
 
   inputs = {
-    nixpkgs.url = "github:nixos/nixpkgs?ref=release-22.11";
+    nixpkgs.url = "github:nixos/nixpkgs/release-25.05";
```
angerman, line 5: Great! I keep being confused about which the right format is though.

angerman (approved):

flake.nix
johnalotoski: AFAIK, there isn't a right or wrong, I just tend to prefer the path style unless it is an ambiguous case, like a branch and a tag with the same name where the query style would disambiguate.

PR #598, opened 2025-05-08
Add mithril network ancillary verification keys

johnalotoski:
Adds mithril network ancillary verification keys:
* The ancillary verification keys are a breaking change required to pull ancillary mithril state (ie: ledger and volatile state) in addition to immutable state.  In earlier mithril releases, the ancillary state was bundled together with the immutable state.  The mithril [v2517.1 release](https://github.com/input-output-hk/mithril/releases/tag/2517.1) and recent [blog article](https://mithril.network/doc/dev-blog/2025/05/06/client-breaking-change/) discuss this change further.
* These keys are sourced from the [Mithril Network configurations page](https://mithril.network/doc/manual/getting-started/network-configurations) at the given Ancillary verification key  url

### DIFF
--- a/cardano-lib/default.nix
+++ b/cardano-lib/default.nix
@@ -145,6 +145,7 @@ let
       smashUrl = "https://smash.cardano-mainnet.iohk.io";
       metadataUrl = "https://tokens.cardano.org";
       mithrilAggregatorEndpointUrl = "https://aggregator.release-mainnet.api.mithril.network/aggregator";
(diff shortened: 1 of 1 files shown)


angerman:
@johnalotoski why are two of them longer than the third one? And where are they used?

johnalotoski:
@angerman -- the key strings are different length due to encoding which is a "hex encoded version of the 32 bytes of the Ed25519 verification key represented as a list of integer values".  When decoded, both vkeys are 32 bytes.

angerman (approved):

PR #601, opened 2025-06-13
Use ouroboros-network defaults for some config

johnalotoski:
* Use ouroboros-network defaults for deadline and sync target related config per network team request
* Update peer snapshots per env
* Update useLedgerAfterSlot per env


crocodile-dentist (approved):
lgtm re using default exported peer selection targets from ouroboros-network with a minor override for BP

angerman (approved):

PR #602, opened 2025-07-25
WIP 2025-07-25

johnalotoski:
* WIP


cardano-lib/testnet-template/config-legacy.json
```diff
@@ -76,7 +76,7 @@
   ],
   "hasEKG": 12788,
   "hasPrometheus": [
-    "0.0.0.0",
+    "127.0.0.1",
```
angerman, line 79: Why this change?

cardano-lib/testnet-template/config.json
```diff
@@ -5,7 +5,6 @@
   "ByronGenesisFile": "byron-genesis.json",
   "ConsensusMode": "PraosMode",
   "ConwayGenesisFile": "conway-genesis.json",
-  "EnableP2P": true,
```
angerman, line 8: Just a random thought as I'm reading this again: These random changes in the flags put me off every time, am I the only one? Some versioning would be nice, it could potentially also allow the node to provide better error messages.

angerman (approved):

cardano-lib/testnet-template/config-legacy.json
johnalotoski: For the standard networks, metrics are exported by default to localhost so as to not expose them to the world if a suitable firewall isn't already in place.  This testnet config was inconsistent with the other cfgs and is now made consistent with them.  Arguably this is less important for a testnet template config, but overall still seemed reasonable to make consistent in approach.

cardano-lib/testnet-template/config.json
johnalotoski: Overall, the more changes in flags we have the more it can potentially cause confusion, so yes I agree that flag changes would ideally be minimized, well documented and have helpful logging as well.  We do provide a `minNodeVersion` in the config to indicate the minimum node version the bundled configs with any new changes are compatible with.  The release notes should indicate this change; I'll communicate the suggestion for logging also.

PR #611, opened 2026-04-16
Fix darwin code signatures in setGitRev using darwin.signingUtils

angerman:
## Summary

- The `setGitRev` function uses `set-git-rev` to patch binary content in Haskell executables, embedding git revision hashes.
- On aarch64-darwin (Apple Silicon), this binary patching invalidates the ad-hoc code signature that the linker originally created. Apple Silicon requires all executables to have valid code signatures -- without re-signing, the patched binary is killed on launch.
- This PR uses `darwin.signingUtils` and its `signIfRequired` wrapper (matching the pattern from cardano-node's `nix/set-git-rev.nix`) to properly re-sign executables after patching on darwin platforms.

## Changes

In `overlays/haskell-nix-extra/default.nix`:

1. Added `nativeBuildInputs = optionals stdenv.hostPlatform.isDarwin [ buildPackages.darwin.signingUtils ];` to bring `signIfRequired` into scope (sourced from `buildPackages` to match the `buildPackages.runCommand` context and support cross-compilation).
2. Added a post-patching loop that calls `signIfRequired` on each executable to re-sign after `set-git-rev` invalidates the original code signature.

This is preferable to a raw `/usr/bin/codesign` call because `darwin.signingUtils` handles signing correctly within the Nix sandbox and supports both ad-hoc and identity-based signing as needed by the nixpkgs infrastructure.

## Test plan

- [ ] Build a Haskell package that uses `setGitRev` on aarch64-darwin and verify the resulting executable runs without code signature errors
- [ ] Verify the fix is a no-op on linux (no codesign invocation, no `darwin.signingUtils` in `nativeBuildInputs`)

### DIFF
--- a/overlays/haskell-nix-extra/default.nix
+++ b/overlays/haskell-nix-extra/default.nix
@@ -55,7 +55,8 @@ in final: prev: with final; with lib; {
         passthru = drv.passthru // (optionalAttrs (drv.passthru ? exeName) {
           exePath = newdrv + "/bin/${drv.passthru.exeName}";
         });
-      } ''
+        nativeBuildInputs = optionals stdenv.hostPlatform.isDarwin [ darwin.signingUtils ];
+      } (''
           mkdir -p $out
           # We link rather than copy from original, to save some space/time:
           ln -s ${drv}/* $out/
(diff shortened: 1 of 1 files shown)


hamishmack (approved):
LGTM

